PR #6, opened 2022-10-31
Create pubspec.yaml

polina-c:
Tests are expected to fail while there is no code yet.


kenzieschmoll:
Can you add an empty test so that the CI passes?

PR #7, opened 2022-11-01
Copy model from devtools to leak_tracker.

polina-c:
The model code is copied from [devtools](https://github.com/flutter/devtools/tree/master/packages/devtools_app/lib/src/screens/memory/panes/leaks/instrumentation) in order to be cleared from devtools by referencing leak_tracker.

### DIFF
--- a/lib/leak_tracker.dart
+++ b/lib/leak_tracker.dart
@@ -1,0 +1,5 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'src/model.dart';
--- a/lib/leak_tracker.dart
+++ b/lib/leak_tracker.dart
@@ -1,3 +1,6 @@
 // Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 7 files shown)


lib/leak_tracker.dart
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
```
CoderDake, line 3: Is this empty file necessary?

lib/src/_util.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// Compares two maps for element-by-element equality.
+///
+/// Copied from
+/// https://github.com/flutter/flutter/blob/37b72342b0ce86fbfc238a9d43e524608b89af3a/packages/flutter/lib/src/foundation/collections.dart#L22
+bool mapEquals<T, U>(Map<T, U>? a, Map<T, U>? b) {
```
CoderDake, line 9: The unit tests should probably be copied as well.

lib/leak_tracker.dart
polina-c: It is a placeholder for now. Created it, to have it reviewed, as the file structure is also subject for review. 

Added comment.

CoderDake (approved):

PR #11, opened 2022-11-09
Enable flutter testing.

polina-c:


.github/workflows/flutter.yaml
```diff
@@ -0,0 +1,38 @@
+name: Dart
```
devoncarew, line 1: The name of this file is not critical; some repos call it `dart.yaml`, some `build.yaml` / `ci.yaml`. Whatever you use, I'd make sure that it agrees with the CI badge at the top of the readme (https://raw.githubusercontent.com/dart-lang/leak_tracker/main/README.md).

.github/workflows/flutter.yaml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: clone the repo
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8
+
+      - name: install Flutter sdk
+        uses: subosito/flutter-action@dbf1fa04f4d2e52c33185153d06cdb5443aa189d
+        with:
+          channel: 'beta'
+
+      - name: content
```
devoncarew, line 25: You might want something more descriptive for the step title (this is printed out when the action runs).

I don't know that you need `pwd` or `ls` below (I imagine this was here for debugging the workflow?); they are however not hurting anything.

analysis_options.yaml
```diff
@@ -12,6 +12,7 @@ analyzer:
     missing_return: warning
   exclude:
     - build/**
+    - flutter_packages/**
```
devoncarew, line 15: I'm curious why this is excluded (is this dir being created somewhere)?

test/flutter/flutter_sdk_test.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test('Flutter tests function for this folder.', () {
```
devoncarew, line 8: 🎉

devoncarew (approved):
lgtm, though I would do a little cleanup in the .github/workflows/flutter.yaml file before landing, and would definitely either rename the file, or update the reference to it from the readme.

.github/workflows/flutter.yaml
CoderDake, line 1: We may need copyrights at the top of these files too.

CoderDake (approved):
not much to add on top of Devon :)

.github/workflows/flutter.yaml
polina-c: And fixed the reference. I did not realize it is referenced. That's cool!

.github/workflows/flutter.yaml
polina-c: Cleaned up.

analysis_options.yaml
polina-c: Yes, it is legacy from removed step. Removed. Thanks.

.github/workflows/flutter.yaml
polina-c: We do not have the in other workflows in dart repo: https://github.com/dart-lang/linter/blob/main/.github/workflows/build.yaml

PR #18, opened 2022-11-23
Add leak checker.

polina-c:


test/dart/_leak_checker_test.dart
```diff
@@ -0,0 +1,248 @@
...
+  setUp(() {
+    leakProvider = _MockLeakProvider();
+
+    stdout = _MockStdoutSink();
+    devtools = _MockDevToolsSink();
+    listened = _ListenedSink();
+  });
+
+  test('Mocks emulate production well.', () {
+    _SummaryValues.verifyValues();
```
CoderDake, line 75: it would be good to pull the contents of this function into the test so it is immediately apparent what is being tested

test/dart/_leak_checker_test.dart
```diff
@@ -0,0 +1,248 @@
...
+    await Future.delayed(doublePeriod);
+    expect(stdout.store, isEmpty);
+    expect(devtools.store, isEmpty);
+    expect(listened.store, isEmpty);
+
+    // Check leaks and make sure they signaled one time.
+    checker.checkLeaks();
+    stdout.checkAndClear(_SummaryValues.nonZero);
+    devtools.checkAndClear(_SummaryValues.nonZero);
+    listened.checkAndClear(_SummaryValues.nonZero);
```
CoderDake, line 202: expectations inside helper methods can make things a bit less clear,
It might be better to have an optional getter to get the values and do the expectation inside the test.

CoderDake (approved):

test/dart/_leak_checker_test.dart
polina-c: I am not sure I understood your suggestion. If I just pull the method content to the test, it will make the test much bigger and less readable.
Renamed the method for now to make it more understandable.

CoderDake: For example for each of these it would be clearer to state what is being tested in the test, not hiding it inside the helper :)
 
 it is a bit of a nit pick, but generally this is a practicer that is good to shoot for
    ```
    expect(listened.store, hasLength(1));
    expect(listened.store.first.matches(summary), isTrue);
    listened.clear();
    ```

polina-c: it will make it harder to read. how about passing array instead of one item? it should make it clear that the method validates the content of the array.

CoderDake: I like that compromise for sure!

polina-c: Applied in https://github.com/dart-lang/leak_tracker/pull/20

PR #19, opened 2022-11-25
Update API and add integration testing.

polina-c:


lib/src/_leak_checker.dart
```diff
@@ -46,6 +46,9 @@ class LeakChecker {
   final LeakProvider leakProvider;
 
   void checkLeaks() {
+    // TODO: remove after debugging
+    print(StackTrace.current);
```
polina-c, line 50: remove before merging

test/flutter/integration/README.md
```diff
@@ -0,0 +1,2 @@
+This folder tests non-mocked public API of leak tracker.
+The tests can serve as exapmles for regression leak-testing for Flutter widgets.
```
CoderDake, line 2:
```suggestion
The tests can serve as examples for regression leak-testing for Flutter widgets.
```

test/test_infra/helpers/gc.dart
```diff
@@ -0,0 +1,21 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:developer';
+
+/// Forces garbage collection by aggressive memory allocation.
+Future<void> forceGC() async {
+  final _storage = <List<DateTime>>[];
+
+  void allocateMemory() {
+    _storage.add(Iterable.generate(10000, (_) => DateTime.now()).toList());
```
CoderDake, line 12: If we are running these stress tests on our github bots, we may need to double check if we will be taxing any shared quotas.
Do you know if this is something that will take a lot of resources on the bots, or use a lot of time units?

lib/src/leak_tracker_model.dart
```diff
@@ -16,6 +16,16 @@ class LeakTrackingConfiguration {
     this.classesToCollectStackTraceOnDisposal = const {},
   });
 
+  /// The leak tracker will not auto check leaks, and, when
+  /// leak checking is invoked, will notify only [listener].
```
CoderDake, line 20:
```suggestion
  /// leak checking is invoked, will notify only [leakListener].
```

test/dart/platform_test.dart
```diff
@@ -0,0 +1,51 @@
...
+  return WeakReference(theObject);
+}
+
+late List<List<DateTime>> _storage;
+
+void _allocateMemory() {
+  _storage.add(Iterable.generate(10000, (_) => DateTime.now()).toList());
+}
+
+/// Tests for non-guaranteed asuumptions about Dart garbage collector,
```
CoderDake, line 28:
```suggestion
/// Tests for non-guaranteed assumptions about Dart garbage collector,
```

CoderDake (approved):
LGTM, but I do wonder if the stress tests put any undue stress on the github bots, or our shared resources.
Do you know if that will be an issue?

test/test_infra/helpers/gc.dart
polina-c: If everything is as expected, this should not take a lot of resources: the cycle will stop at first garbage collection, that should happen much earlier than any significant impact for bots.

polina-c:
> LGTM, but I do wonder if the stress tests put any undue stress on the github bots, or our shared resources. Do you know if that will be an issue?

If everything is as expected, this should not take noticeable resources: the cycle will stop at first garbage collection, that should happen much earlier than any significant impact for bots.

PR #20, opened 2022-11-29
Integrate with DevTools.

polina-c:


test/dart/devtools_integration/_envelopes_test.dart
```diff
@@ -0,0 +1,33 @@
...
+    expect(codesInEnvelopes, hasLength(Codes.values.length));
+  });
+
+  test('envelopes are unique by type', () {
+    final types = Set.from(envelopes.map((e) => e.type));
+    expect(types, hasLength(envelopes.length));
+  });
+
+  for (final message in messages) {
+    test('envelopes preserve original type', () {
```
CoderDake, line 26:
```suggestion
    test('envelopes preserve $runtimeType type', () {
```
When debugging failing tests it can be helpful to include information about which case is being tested, in the name

lib/src/devtools_integration/_envelopes.dart
```diff
@@ -0,0 +1,150 @@
...
+  static Codes byName(String name) =>
+      Codes.values.where((e) => e.name == name).single;
+}
+
+class _JsonFields {
+  static const envelopeCode = 'code';
+  static const content = 'content';
+}
+
+class Envelope<T> {
```
CoderDake, line 47: Some information about could be helpful here:

- what is an envelope?
- what is it used for?
- how do we expect it to behave?
- any other info that might be helpful

lib/src/devtools_integration/primitives.dart
```diff
@@ -0,0 +1,10 @@
+
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Name of extension to integrate the leak tracker with DevDools.
+const String memoryLeakTrackingExtensionName = 'ext.dart.memoryLeakTracking';
+
+/// Version of protocol, executed by the application.
+const String appLeakTrackerProtocolVersion = '1';
```
CoderDake, line 10: Will this be important for caching? 
It may be worth using semver in case we ever need to release incremental patches/fixes etc.

test/dart/devtools_integration/delivery_test.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import '../../test_infra/data/messages.dart';
+
+void main() {
+  setUp(() {
+    verifyTestsCoverAllEnvelopes();
+  });
+
+  for (final message in messages) {
+    test('delivery classes serialize message', () {
```
CoderDake, line 17:
```suggestion
    test('delivery classes serialize ${message.runtimeType} message', () {
```
It would be good to add some message detail to this message to help with future debugging

test/dart/devtools_integration/delivery_test.dart
```diff
@@ -0,0 +1,45 @@
...
+      final envelope = envelopeByType(message.runtimeType);
+
+      switch (envelope.channel) {
+        case Channel.requestToApp:
```
CoderDake, line 21: Doing a switch statement in unit tests decreases readability.
It might be more readable to make separate tests  where you grab the types you want, then operate on those types.

pubspec.yaml
```diff
@@ -1,14 +1,15 @@
 name: leak_tracker
-version: 1.0.0-dev
+version: 1.0.0-1.0.dev
```
CoderDake, line 2:
```suggestion
version: 1.0.0-dev.1.0
```
Semver rules have the prelease with the name before the pre-release version
https://semver.org/#spec-item-9

lib/src/devtools_integration/primitives.dart
polina-c: I expect it to be simple. For now I want to check version. When we need to introduce breaking changes, we will decide how complicated it should be.

CoderDake (approved):

PR #24, opened 2022-12-08
Add test for pumpWidget.

polina-c:
1. Create integration test for pumpWidget.
2. Make `disposalTimeBuffer` configurable, because it should be zero for unit tests.
3. Add testing helper `withLeakTracking`


test/flutter/integration_test.dart
```diff
@@ -0,0 +1,35 @@
...
+import '../test_infra/data/dart_classes.dart';
+import '../test_infra/data/flutter_classes.dart';
+
+/// Tests for non-mocked public API of leak tracker.
+///
+/// Can serve as examples for regression leak-testing for Flutter widgets.
+void main() {
+  testWidgets('Leaks in pumpWidget are detected.', (WidgetTester tester) async {
+    await tester.runAsync(() async {
+      final leaks = await withLeakTracking(
```
CoderDake, line 17: exposing testing helpers through the package is a bit new to me.
@kenzieschmoll Do you know if this is something that is ok to do?

polina-c: I ended up renaming 'testing' to 'orchestration', because it can be utilized not just in testing. Thanks for raising the concern.

PR #26, opened 2022-12-13
Update documentation for leak tracker.

polina-c:
A lot of content is copied from https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/screens/memory/panes/leaks/LEAK_TRACKING.md.

After the PR submission LEAK_TRACKING.md will be deleted.

After tech review, the text should be reviewed with tech writers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The tool assumes that, with proper memory management,
 an object's disposal and garbage collection should happen sequentially,
 close to each other.
 
-By monitoring disposal and GC events, the tool detects different types of leaks:
+By monitoring disposal and Garbage Collect events, the tool detects different types of leaks:
 
 - **Not disposed, but GCed (not-disposed)**:
     - a disposable object was GCed,
--- a/README.md
+++ b/README.md
@@ -63,17 +63,19 @@ close to each other.
 
 By monitoring disposal and GC events, the tool detects different types of leaks:
 
-**Not disposed, but GCed (not-disposed)**: a disposable object was GCed,
-without being disposed first. This means that the object's disposable content
-is using memory after the object is no longer needed.
-To fix the leak, invoke `dispose()` to free up the memory.
+- **Not disposed, but GCed (not-disposed)**:
+    - a disposable object was GCed,
+       without being disposed first. This means that the object's disposable content
+       is using memory after the object is no longer needed.
+       To fix the leak, invoke `dispose()` to free up the memory.
 
-**Disposed, but not GCed (not-GCed)**: an object was disposed,
-but not GCed after certain number of GC events. This means that
-a reference to the object is preventing it from being
-garbage collected after it's no longer needed.
-To fix the leak, after disposal assign all reachable references
-of the object to null:
+- **Disposed, but not GCed (not-GCed)**:
+    - an object was disposed,
+       but not GCed after certain number of GC events. This means that
+       a reference to the object is preventing it from being
+       garbage collected after it's no longer needed.
+       To fix the leak, after disposal assign all reachable references
+       of the object to null:
 
 ```
 myField.dispose();
--- a/README.md
+++ b/README.md
@@ -82,14 +82,16 @@ myField.dispose();
 myField = null;
 ```
 
-**Disposed and GCed late (GCed-late)**: an object was disposed and then GCed,
-but GC happened later than expected. This means the retaining path was
-holding the object in memory for some period, but then disappeared.
+- **Disposed and GCed late (GCed-late)**:
+    - an object was disposed and then GCed,
+       but GC happened later than expected. This means the retaining path was
+       holding the object in memory for some period, but then disappeared.
 
-**Disposed, but not GCed, without path (not-GCed-without-path)**: an object
-was disposed and not GCed when expected, but retaining path is not detected,
-that means that the object will be most likely GCed in the next GC cycle,
-and the leak will convert to **GCed-late** leak.
+- **Disposed, but not GCed, without path (not-GCed-without-path)**:
+    - an object
+       was disposed and not GCed when expected, but retaining path is not detected,
+       that means that the object will be most likely GCed in the next GC cycle,
+       and the leak will convert to **GCed-late** leak.
 
 ### Culprits and victims
 
--- a/README.md
+++ b/README.md
@@ -56,10 +56,9 @@ Before reading about leak tracking, understand [Dart memory concepts](https://do
 ### Addressed leak types
 
 The leak tracker can catch only certain types of leaks, in particular, related to timing of disposal and garbage collection.
-
-The tool assumes that, with proper memory management,
-an object's disposal and garbage collection should happen sequentially,
-close to each other. I.e. the object should be garbage collected
+With proper memory management, this tool assumes that, 
+an object's disposal and garbage collection occur in quick succession.
+That is, the object should be garbage collected
 during next garbage collection cycle after disposal.
 
 By monitoring disposal and Garbage Collect events, the tool detects different types of leaks:
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ The leak tracker will catch leaks only for instrumented objects (See [concepts](
 However, the good news is:
 
 1. Most of disposable Flutter Framework classes are already instrumented. So, Widget related
-leaks in your Flutter application will be cought without additional effort.
+leaks in your Flutter application will be caught without additional effort.
 
 2. If a leak involves at least one instrumented object, the leak will be caught and all
 other objects, even non-instrumented, will stop leaking as well.
--- a/README.md
+++ b/README.md
@@ -1,16 +1,17 @@
+
+Coming soon! See https://github.com/flutter/devtools/issues/3951.
+
+The text below is under construction.
+
 [![CI](https://github.com/dart-lang/leak_tracker/actions/workflows/ci.yaml/badge.svg)](https://github.com/dart-lang/leak_tracker/actions/workflows/ci.yaml)
 
 # Memory Leak Tracker
 
 This is a framework for memory leak tracking for Dart and Flutter applications.
 
-Coming soon! See https://github.com/flutter/devtools/issues/3951.
+## Quick start to track leaks for Flutter
 
-The text below this sentence is under construction.
-
-## Getting started
-
-To track memory leaks in your Flutter application:
+### Flutter application
 
 1. Before `runApp` invocation, enable leak tracking, and connect the Flutter memory allocation events:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+```
+
+**Disposed and GCed late (GCed-late)**: an object was disposed and then GCed,
+but GC happened later than expected. This means the retaining path was
+holding the object in memory for some period, but then disappeared.
+
+**Disposed, but not GCed, without path (not-GCed-without-path)**: an object
+was disposed and not GCed when expected, but retaining path is not detected,
+that means that the object will be most likely GCed in the next GC cycle,
+and the leak will convert to **GCed-late** leak.
```
CoderDake, line 90:
```suggestion
- **Disposed and GCed late (GCed-late)**:
    - an object was disposed and then GCed,
       but GC happened later than expected. This means the retaining path was
       holding the object in memory for some period, but then disappeared.

- **Disposed, but not GCed, without path (not-GCed-without-path)**:
    - an object
       was disposed and not GCed when expected, but retaining path is not detected,
       that means that the object will be most likely GCed in the next GC cycle,
       and the leak will convert to **GCed-late** leak.
```

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+
+The leak tracker can catch only certain types of leaks, in particular, related to timing of disposal and garbage collection.
+
+The tool assumes that, with proper memory management,
+an object's disposal and garbage collection should happen sequentially,
+close to each other.
+
+By monitoring disposal and GC events, the tool detects different types of leaks:
+
+**Not disposed, but GCed (not-disposed)**: a disposable object was GCed,
```
CoderDake, line 66: It may be good to split the bodies of these enties into "Description" and "Fix" bullets, so that it is easy to visually parse the sections and their value.

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+without being disposed first. This means that the object's disposable content
+is using memory after the object is no longer needed.
+To fix the leak, invoke `dispose()` to free up the memory.
```
CoderDake, line 69: what should the reader call dispose on? it might be helpful to mention that here

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+
+### Addressed leak types
+
+The leak tracker can catch only certain types of leaks, in particular, related to timing of disposal and garbage collection.
+
+The tool assumes that, with proper memory management,
+an object's disposal and garbage collection should happen sequentially,
+close to each other.
+
+By monitoring disposal and GC events, the tool detects different types of leaks:
```
CoderDake, line 64:
```suggestion
By monitoring disposal and Garbage Collect events, the tool detects different types of leaks:
```
Since this is an instructive document, it may help to spell out "Garbage Collect" for GC and "Garbage Collected" from GCed.
We don't have any space constraints here, so I think clarity takes precedence over brevity

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+of the object to null:
+
+```
+myField.dispose();
+myField = null;
+```
+
+**Disposed and GCed late (GCed-late)**: an object was disposed and then GCed,
+but GC happened later than expected. This means the retaining path was
+holding the object in memory for some period, but then disappeared.
```
CoderDake, line 85: Is there a hint or fix we can lead them to here?

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+
+**Disposed, but not GCed, without path (not-GCed-without-path)**: an object
+was disposed and not GCed when expected, but retaining path is not detected,
+that means that the object will be most likely GCed in the next GC cycle,
+and the leak will convert to **GCed-late** leak.
+
+### Culprits and victims
```
CoderDake, line 92: This terminology may end up being a bit too harsh. It may be worth asking the tech writers if they have different analogies they would prefer using here.

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+      l1-->l3;
+      l2-->l4;
+      l3-->l4;
+```
+
+## Limitations
+
+### By tracked classes
+
+The leak traker will catch leaks only for instrumented onjects (See [concepts](#leak-tracking-concepts) for details).
```
CoderDake, line 127:
```suggestion
The leak tracker will catch leaks only for instrumented objects (See [concepts](#leak-tracking-concepts) for details).
```

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+
+However, the good news are:
```
CoderDake, line 129:
```suggestion
However, the good news is:
```

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+
+1. Most of disposable Flutter Framework classes are already instrumented. So, Widget related
+leaks in your Flutter application will be cought without additional effort.
```
CoderDake, line 132:
```suggestion
leaks in your Flutter application will be caught without additional effort.
```

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+
+NOTE: If you are interested in enabling leak tracking for release mode, please, comment [here](https://github.com/dart-lang/leak_tracker/issues/25).
+
+## Instrument your code
+
+If you want to catch leaks for objects outside of Flutter Framework,
+(that are already instrumented) you need to instrument them.
+
+For each tracked object the library should get two signals from your code:
+(1) the object is created and (2) the object is not in use. It is most convenient to give first signal in constractor and the second one in the method `dispose`:
```
CoderDake, line 164:
```suggestion
(1) the object is created and (2) the object is not in use. It is most convenient to give the first signal in the constructor and the second one in the `dispose` method:
```

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+However, the good news are:
+
+1. Most of disposable Flutter Framework classes are already instrumented. So, Widget related
+leaks in your Flutter application will be cought without additional effort.
+
+2. If a leak involves at least one instrumented object, the leak will be caught and all
+other objects, even non-instrumented, will stop leaking as well.
+
+See [the instrumentation guidance](#instrument-your-code).
+
```
CoderDake, line 138: It might be a good idea to add a section mentioning how to report a bug if they notice that a framework entity has a leak

CoderDake (commented):
It may be good to add a section

README.md
polina-c: Tech writers and UXR already reviewed it and they loved it, because it is extremely illustrative :)

README.md
```diff
@@ -29,6 +30,206 @@ runApp(...
...
+    throwOnLeaks: false,
+  );
+
+  expect(leaks.total, 0);
+});
+```
+
+## Leak tracking concepts
+
+Before reading about leak tracking, understand [Dart memory concepts](https://docs.flutter.dev/development/tools/devtools/memory#basic-memory-concepts).
```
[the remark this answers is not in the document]
polina-c: Normally 'Quick start' means the person does not want to learn anything, they just wants to make it working. So, I would leave it here. 

But, tech writers may agree with you, so leaving the comment unresolved.

README.md
polina-c: Thanks. Updated the sentence to show it is most likely in-app problem.

polina-c:
@atsansone or @sfshaza2 , can you review?

README.md
```diff
@@ -59,40 +59,48 @@ The leak tracker can catch only certain types of leaks, in particular, related t
...
+    - **Fix**: the same as for **not-GCed**
+
 - **Disposed, but not GCed, without path (not-GCed-without-path)**:
-    - an object
+    - **Definition**: an object
        was disposed and not GCed when expected, but retaining path is not detected,
        that means that the object will be most likely GCed in the next GC cycle,
        and the leak will convert to **GCed-late** leak.
 
+    - **Fix**: please, [create issue](https://github.com/dart-lang/leak_tracker/issues) if you see this type of leaks, as it means something is wrong with the tool.
```
atsansone, line 102: **issue:** This information would be better presented as a table. I would replace it with the following:
```suggestion
| Memory Leak Type          | Disposed | Garbage Collected | Explanation                                                                                                                                                                                                                                   | Fix                                                                                                                              |   |
|---------------------------|:--------:|:-----------------:|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------|---|
| **not-disposed**          |    No    |        Yes        | The object's content continues to use memory after the program no longer needs the object.                                                                                                                                                    | Call `dispose()` to free up the memory.                                                                                          |   |
| **not-GCed**              |    Yes   |         No        | Memory disposed of an object, but did not free the memory that object occupied after certain number of GC events. A reference to the object prevents removing it from memory after the program no longer needs it to be in memory.            | Assign all reachable references of the object to `null` after disposal.  ```dart myField.dispose(); myField = null; ```          |   |
| **GCed-late**             |    Yes   |   Yes, but late   | Memory disposed of an object, then cleared the object from memory later than expected. The retaining path continued to hold the object in memory for some period, but then disappeared.                                                       | Assign all reachable references of the object to `null` after disposal.  ```dart myField.dispose(); myField = null; ```          |   |
| **not-GCed-without-path** |    Yes   |  No, without path | Memory disposed of an object and did not clear the object from memory when expected, but memory cannot detect the object's retaining path. The object should be cleared in the next GC cycle. The leak then converts to a **GCed-late** leak. | [Create an issue](https://github.com/dart-lang/leak_tracker/issues). If you see this type of leak, the tool might have an issue. |   |
```

README.md
```diff
@@ -132,8 +140,9 @@ The leak tracker will catch leaks only for instrumented objects (See [concepts](
 
 However, the good news is:
 
-1. Most of disposable Flutter Framework classes are already instrumented. So, Widget related
-leaks in your Flutter application will be caught without additional effort.
+1. Most of disposable Flutter Framework classes are already instrumented. So, if
+your Flutter application manages widgets in a way that creates leaks, the leaks
+will be caught without additional effort.
```
atsansone, line 145: **issue:** This reads with some difficulty. I would recommend rephrasing somewhat.
```suggestion
1. Most disposable Flutter Framework classes include instrumentation.
If how your Flutter app manages widgets results in leaks, Flutter will catch them.
```

atsansone (changes requested):
@polina-c : Recommending a style guide correction, changing presentation to a table, and simplifying some other language.

atsansone (approved):

README.md
polina-c: I love how the table look, but the table format makes it very hard to edit and review the page. So, leaving it bulleted list for the sake of maintainability.

README.md
```diff
@@ -27,8 +29,244 @@ runApp(...
...
+
+```dart
+test('...', () async {
+  final leaks = await withLeakTracking(
+    () async {
+      ...
+    },
+  );
+
+  expect(leaks, leakFree);
```
CoderDake, line 49: What is `leaksFree`? I can't seem to find a reference to it in the repo

CoderDake (approved):

README.md
polina-c: Will show up in the next PR

PR #27, opened 2022-12-13
Disable leak_tracker for release mode.

polina-c:


CoderDake (approved):
super cool. Learned something new today about using assert sneaking in when only in an assertable mode.

PR #28, opened 2022-12-15
Tests and fixes.

polina-c:
- Add test for non-disposed member
- Improve YAML formatting to format call-stack nicely (see description in https://github.com/flutter/flutter/issues/117063)
- When tracking is off, no-op silently for object events, instead of failing
- Other minor fixes


lib/src/_util.dart
```diff
@@ -31,3 +31,8 @@ T cast<T>(value) {
     '$value is of type ${value.runtimeType} that is not subtype of $T',
   );
 }
+
+Object? safeToEncodable(Object? object) {
```
CoderDake, line 35:
```suggestion
Object? toEncodable(Object? object) {
```
This name reads a bit weird
Could this just be left as toEncodable? Otherwise a JSDoc may be nice here giving some context.

lib/src/orchestration.dart
```diff
@@ -17,38 +20,72 @@ class MemoryLeaksDetectedError extends StateError {
 
 /// Tests the functionality with leak tracking.
 ///
-/// If invoked in tests inside `testWidget`, use `tester.runAsync`.
+/// If you use it inside `testWidget`, connect Flutter objects and use `tester.runAsync`:
```
CoderDake, line 23:
```suggestion
/// If you use `withLeakTracking` inside `testWidget`, connect Flutter objects and use `tester.runAsync`:
```

test/test_infra/data/flutter_classes.dart
```diff
@@ -22,3 +22,7 @@ class StatelessLeakingWidget extends StatelessWidget {
     return const Placeholder();
   }
 }
+
+class ValueNotifierNotDisposer {
```
CoderDake, line 26: JSDoc would be good here explaining what this is and why it might be used.

CoderDake (approved):

PR #31, opened 2022-12-21
Add option to reset leak tracking.

polina-c:


README.md
```diff
@@ -260,7 +260,7 @@ for troubleshooting:
 1. By passing `stackTraceCollectionConfig`
 to `withLeakTracking` or `enableLeakTracking`.
 
-<video src="md_images/collect_callstack.mov" controls="controls" playsinline="playsinline"></video>
+https://user-images.githubusercontent.com/12115586/208321882-ecb96152-3aa7-4671-800e-f2eb8c18149e.mov
```
CoderDake, line 263: Does this automatically show in the markdown? or do we need to turn it into a clickable link?

lib/src/leak_tracker.dart
```diff
@@ -24,11 +24,16 @@ ObjectTracker _theObjectTracker() {
 ///
 /// The leak tracking will function only for debug/profile/developer mode.
 /// See usage guidance at https://github.com/dart-lang/leak_tracker.
```
CoderDake, line 26: JSDOC info should be added for the new param

lib/src/leak_tracker.dart
```diff
@@ -24,11 +24,16 @@ ObjectTracker _theObjectTracker() {
...
-void enableLeakTracking({LeakTrackingConfiguration? config}) {
+void enableLeakTracking({
+  LeakTrackingConfiguration? config,
+  bool resetIfEnabled = false,
```
CoderDake, line 29: This name is a bit confusing, I don't get an idea of what is accomplished when it is set.

lib/src/leak_tracker.dart
```diff
@@ -24,11 +24,16 @@ ObjectTracker _theObjectTracker() {
...
+}) {
   assert(() {
     final theConfig = config ??= const LeakTrackingConfiguration();
-    if (_objectTracker.value != null)
-      throw StateError('Leak tracking is alredy enabled.');
+    if (_objectTracker.value != null) {
+      if (!resetIfEnabled) throw StateError('Leak tracking is alredy enabled.');
```
CoderDake, line 34: Is there a reason why we allow both behaviours? Could we just always reset ?

README.md
polina-c: Original way did not work, but this does. and I do not know why. See here: https://github.com/dart-lang/leak_tracker/blob/e4edef3cfd431f8c438e6353194da0eba197e118/README.md

lib/src/leak_tracker.dart
polina-c: Normally leak tracking should be set once for application. So, the second one is a smell.

Reset is needed just for tests.

lib/src/leak_tracker.dart
polina-c: Renamed to resetIfAlreadyEnabled. Does it help?

lib/src/leak_tracker.dart
CoderDake, line 34:
```suggestion
      if (!resetIfEnabled) throw StateError('Leak tracking is already enabled.');
```

CoderDake (approved):

PR #37, opened 2023-01-11
Update package:vm_service dependency to >=9.0.0 <12.0.0

derekxu16:


CoderDake:
It looks like there are some dependency issues we will need to address before this can be merged

derekxu16:
Yeah I'm still rolling the `vm_service` version bump through some other packages.

polina-c:
Let's wait when v10 is available for flutter master

polina-c (approved):

PR #39, opened 2023-01-26
Improve documentation.

polina-c:


polina-c:
Loaded picture here to link as picture in readme, so that it is available in pub.dev description too.

Screenshot 2023-01-26 at 4 31 54 PM

CoderDake (approved):

PR #41, opened 2023-03-22
Add auto updater for diagrams.

polina-c:
This PR deletes the diagrams to verify they will be regenerated.


CONTRIBUTING.md
```diff
@@ -49,3 +49,10 @@ This project follows
 
 We pledge to maintain an open and welcoming environment. For details, see our
 [code of conduct](https://dart.dev/code-of-conduct).
+
+## Regenerate Dependency Diagrams
+
+The [LayerLens](https://github.com/polina-c/layerlens) diagrams
+will be auto-regenerared by GitHub after eash push to main branch.
```
CoderDake, line 56:
```suggestion
will be auto-regenerated by GitHub after each push to main branch.
```

.github/workflows/generate_diagrams.yaml
```diff
@@ -0,0 +1,32 @@
...
+      run: dart --version
+
+    - name: dart pub get
+      run: dart pub get
+
+    - name: generate
+      run: dart run layerlens
+
+    - name: Commit files
+      uses: stefanzweifel/git-auto-commit-action@v4
```
CoderDake, line 30: It looks like the default committer will be `github-actions[bot]`
I think the CLA workflow will end up failing for those commits. So you may need to tweak this setting to accomodate that.a

CoderDake (approved):

.github/workflows/generate_diagrams.yaml
polina-c: Let's try it out. We will revert the change, if it does not work.

PR #45, opened 2023-04-13
Improve documentation.

polina-c:


lib/src/orchestration.dart
```diff
@@ -24,20 +24,14 @@ class MemoryLeaksDetectedError extends StateError {
   final Leaks leaks;
 }
 
-/// Tests the functionality with leak tracking.
+/// Runs [callback] with memory leak detection.
 ///
-/// Wrap code inside your test with this method to catch memory
-/// leaks.
+/// See https://github.com/dart-lang/leak_tracker for memory leak definition.
```
goderbauer, line 29: Personally, I don't like documentation that links me to an unspecific page like that and makes me search for the actual definition of a leak. 🤷‍♂️

goderbauer (approved):
LGTM

PR #46, opened 2023-04-18
Test config and flutter lints.

polina-c:
Test config is needed to address the comment: https://github.com/flutter/flutter/pull/124435#discussion_r1169056964

### DIFF
--- a/lib/src/devtools_integration/_protocol.dart
+++ b/lib/src/devtools_integration/_protocol.dart
@@ -68,7 +68,7 @@ Object openEnvelope(
   Map<String, dynamic> message,
   Channel channel,
 ) {
-  final envelope = envelopeByCode(message[_JsonFields.envelopeCode] as String);
+  final envelope = _envelopeByCode(message[_JsonFields.envelopeCode] as String);
   assert(envelope.channel == channel);
   return envelope.decode(message[_JsonFields.content]);
 }
@@ -151,20 +151,13 @@ final envelopes = [
   ),
 ];
 
-_Envelope<T> envelopeByCode<T>(String codeString) {
+_Envelope<T> _envelopeByCode<T>(String codeString) {
   return _envelopesByCode[codeString]! as _Envelope<T>;
 }
 
+// ignore: library_private_types_in_public_api, pubic for testing
 _Envelope envelopeByType(Type type) => _envelopesByType[type]!;
 
-late final _envelopesByCode = Map<String, _Envelope>.fromIterable(
-  envelopes,
-  key: (e) => (e as _Envelope).code.name,
-  value: (e) => e,
-);
+final _envelopesByCode = {for (var e in envelopes) e.code.name: e};
 
-late final _envelopesByType = Map<Type, _Envelope>.fromIterable(
-  envelopes,
-  key: (e) => e.type,
-  value: (e) => e,
-);
+final _envelopesByType = {for (var e in envelopes) e.type: e};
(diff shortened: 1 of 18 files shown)


lib/src/devtools_integration/_protocol.dart
```diff
@@ -151,20 +151,13 @@ final envelopes = [
   ),
 ];
 
-_Envelope<T> envelopeByCode<T>(String codeString) {
+_Envelope<T> _envelopeByCode<T>(String codeString) {
   return _envelopesByCode[codeString]! as _Envelope<T>;
 }
 
+// ignore: library_private_types_in_public_api, pubic for testing
```
bkonyi, line 158: Nit: `public for testing`

lib/src/leak_tracker_model.dart
```diff
@@ -86,11 +93,42 @@ class LeakTrackingConfiguration {
...
+  final LeakSummaryCallback? onLeaks;
 
   /// Time to allow the disposal invoker to release the reference to the object.
   ///
   /// The default value is pessimistic assuming that user will want to
   /// detect leaks not more often than a second.
   final Duration disposalTimeBuffer;
 }
+
+/// Configuration for leak trecking in unit tests.
```
bkonyi, line 105: Nit: `leak tracking`

lib/src/leak_tracker_model.dart
```diff
@@ -86,11 +93,42 @@ class LeakTrackingConfiguration {
...
+///
+/// The configuration is needed only for test debugging,
+/// not for regular test run.
```
bkonyi, line 108: Nit: `not for regular test runs`

lib/src/leak_tracker_model.dart
```diff
@@ -86,11 +93,42 @@ class LeakTrackingConfiguration {
...
+  const LeakTrackingTestConfig({
+    this.stackTraceCollectionConfig = const StackTraceCollectionConfig(),
+    this.onLeaks,
+    this.failTestOnLeaks = true,
+  });
+
+  /// When to collect stack trace information.
+  ///
+  /// You may need to know call stack to troubleshoot memory leaks.
+  /// Custonize this parameter to collect stack traces when needed.
```
bkonyi, line 120: I'd update this documentation to avoid using the second person. Maybe something like this:

```
When to collect stack trace information.

Customize this parameter to determine when stack traces are collected.
```

lib/src/leak_tracker_model.dart
```diff
@@ -86,11 +93,42 @@ class LeakTrackingConfiguration {
...
+
+  /// Handler to obtain details about collected leaks.
+  ///
+  /// Use the handler to process the collected leak
+  /// details programmatically.
+  final LeaksCallback? onLeaks;
+
+  /// If true, the test will fail if leaks are found.
+  ///
+  /// Set it to false if you want the test to pass, in order
```
bkonyi, line 131: Nit: `If false, the test will not fail if leaks are found to allow for analyzing leaks after the test completes.`

lib/src/devtools_integration/_registration.dart
```diff
@@ -51,15 +54,18 @@ bool setupDevToolsIntegration(
         UnexpectedRequestTypeError(request.runtimeType),
       ).toServiceResponse();
     } catch (error, stack) {
+      // ignore: avoid_print
       print(
```
bkonyi, line 58: Nit: can we use `dart:io` here? If so, consider converting `print` to `stdout.writeln` and `stderr.writeln` where appropriate.

polina-c: Leak tracker will be used eventually for web. Made output a utility method.

PR #53, opened 2023-04-27
Add auto-snapshotting.

polina-c:
https://user-images.githubusercontent.com/12115586/234953319-6f864d25-9854-4126-b4d6-8e114b9045ff.mov


README.md
```diff
@@ -7,291 +7,8 @@ The text below is under construction.
 
 # Memory Leak Tracker
 
-This is a framework for memory leak tracking for Dart and Flutter applications.
+This is a framework for memory issues detection in Dart and Flutter applications. It enables:
```
kenzieschmoll, line 10: nit: "...for detecting memory issues in..."

documentation/AUTOSNAPSHOTTING.md
```diff
@@ -0,0 +1,29 @@
...
+
+TODO (polina-c): add link to DEvTools help section.
+
+## Configure auto-snapshotting
+
+Use the function `autoSnapshotOnMemoryOveruse` to configure auto-snapshotting. We
+recommend to have autosnapshotting off by default, with possibility
+to enable it via command line arguments in case of memory issues.
+
+See example in
```
kenzieschmoll, line 19: missing link to example?

documentation/AUTOSNAPSHOTTING.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+## Lumitations
+
+**Enabled for:** Dart debug and release modes (that are Flutter
+debug and profile modes).
+**Disabled for:** Dart product mode (that is Flutter release mode).
```
kenzieschmoll, line 26: I would list flutter first since the majority of dart users are flutter users

lib/src/autosnapshotting/_snapshot.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import 'model.dart';
+
+/// Saves a memory heap snapshot of the current process, to the [folder].
+///
+/// Returns the name of the file where the snapshot was saved.
+///
+/// The file name contains process id, snapshot number and current RSS.
+SnapshotRecord saveSnapshot(
+  String folder, {
```
kenzieschmoll, line 18: nit: rename as `directory` to be consistent with dart io terminology?

documentation/AUTOSNAPSHOTTING.md
```diff
@@ -10,20 +10,21 @@ The snapshots can be later analysed in DevTools.
...
+See [example](../more_examples/autosnapshotting/).
 
-## Lumitations
+## Limitations
 
-**Enabled for:** Dart debug and release modes (that are Flutter
-debug and profile modes).
-**Disabled for:** Dart product mode (that is Flutter release mode).
+* **Enabled for:** Flutter
+debug and profile modes (that are Dart debug and release modes).
```
kenzieschmoll, line 26: r/that are/equivalent to

here and below

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import '_snapshot.dart';
+import 'model.dart';
+
+Timer? _theTimer;
```
kenzieschmoll, line 11: nit: _timer

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+final _takenSnapshots = <SnapshotRecord>[];
+late AutoSnapshottingConfig _config;
+bool _snapshottingIsInProgress = false;
+
+/// Enables auto-snapshotting, based on the value of ProcessInfo.currentRss (dart:io).
```
kenzieschmoll, line 16: wrap in [] : [ProcessInfo.currentRss]

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+///
+/// If autosnapshotting is already enabled, resets it.
```
kenzieschmoll, line 18: nit: auto-snapshotting to be consistent with other comments

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+  _config = config;
+  _theTimer = Timer.periodic(config.interval, (_) {
+    if (_snapshottingIsInProgress) return;
+    _snapshottingIsInProgress = true;
+    _maybeTakeSnapshot();
+    _snapshottingIsInProgress = false;
+  });
+}
+
+void _createFolderIfNotExists(String folder) {
```
kenzieschmoll, line 36: nit: r/Folder/Directory. method name and parameter name

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+  final dir = Directory(folder);
+  if (!dir.existsSync()) {
+    dir.createSync(recursive: true);
+  }
+}
+
+extension _SizeConversion on int {
+  int mbToBites() => this * 1024 * 1024;
```
kenzieschmoll, line 44: sp: Bytes

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+}
+
+bool _isFolderOversized() {
+  final folderSize = Directory(_config.folder)
```
kenzieschmoll, line 48: I haven't tried this myself, but can you use the `stat` function to get the directory size? https://api.dart.dev/stable/2.19.6/dart-io/FileSystemEntity/stat.html

lib/src/autosnapshotting/autosnapshotting.dart
```diff
@@ -0,0 +1,111 @@
...
+  return folderSize >= _config.folderSizeLimitMb.mbToBites();
+}
+
+void _maybeTakeSnapshot() {
+  final rss = ProcessInfo.currentRss;
+  if (rss < _config.thresholdMb.mbToBites()) {
+    return;
+  }
+
+  // Folder size validation is havier than rss check, so we do it after.
```
kenzieschmoll, line 62: sp: heavier

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:path/path.dart' as path;
+
+/// A record of a taken snapshot.
+class SnapshotRecord {
```
kenzieschmoll, line 8: maybe `SnapshotMetadata` since "record" is now a loaded term in Dart? Or just `Snapshot`?

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+  final String fileName;
+  final int snapshotNumber;
+  final int rss;
+  final DateTime timestamp;
+}
+
+/// A callback that is called when a snapshot is taken.
+typedef SnapshotCallback = void Function(SnapshotRecord record);
+
+/// Configures auto-snapshotting, based on the value of ProcessInfo.currentRss (dart:io).
```
kenzieschmoll, line 25: wrap with []: [ProcessInfo.currentRss]

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+///
+/// The snapshots will be taken as soon as the value becomes more than [thresholdMb],
```
kenzieschmoll, line 27: docs nit for clarity: "Automatic snapshots will begin to be taken when the rss value is exceeds [threshholdMb]. Snapshots will be saved to [directory]."

Can you also add more explanation for [directory]? What is this path relative to?

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+/// and saved to the [folder]. The snapshots will be re-taken when the value
+/// increases more than by [stepMb], till hitting the [folderSizeLimitMb].
```
kenzieschmoll, line 29: "until the size of [directory] exceeds [directorySizeLimitMb]"

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+///
+/// The [onSnapshot] callback is called when a snapshot is taken.
+///
+/// Set [minDelayBetweenSnapshots] to make sure snapshots do not trigger each other.
+class AutoSnapshottingConfig {
+  const AutoSnapshottingConfig({
+    this.thresholdMb = 1024, // 1Gb
+    this.stepMb = 512, // 0.5Gb
+    this.folder = 'dart_momory_snapshots',
+    this.folderSizeLimitMb = 10000, // 10Gb
```
kenzieschmoll, line 47: nit: replace all uses of folder with directory for consistency
also spelling: memory

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+typedef SnapshotCallback = void Function(SnapshotRecord record);
+
+/// Configures auto-snapshotting, based on the value of ProcessInfo.currentRss (dart:io).
+///
+/// The snapshots will be taken as soon as the value becomes more than [thresholdMb],
+/// and saved to the [folder]. The snapshots will be re-taken when the value
+/// increases more than by [stepMb], till hitting the [folderSizeLimitMb].
+/// The [folder] will be created if it does not exist.
+///
+/// Set [interval] to customize how often to verify memory consumption.
```
kenzieschmoll, line 32: I don't understand the difference between interval and minDelayBetweenSnapshots. Can you make that more clear in the dart doc since this will be a public API users read?

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+///
+/// If [stepMb] is null, only one snapshot will be taken.
+///
+/// The folder size will be chaecked before saving snapshot, so the folder
```
kenzieschmoll, line 36: sp: checked

"The directory size will be checked before a snapshot is taken and saved, so the directory may exceed the size specified by [directorySizeLimitMb] depending on the size of the snapshot."

lib/src/autosnapshotting/model.dart
kenzieschmoll, line 47: shouldn't this default value be 10240 to be 10gb?

lib/src/autosnapshotting/model.dart
```diff
@@ -0,0 +1,62 @@
...
+    this.onSnapshot,
+  });
+
+  final int thresholdMb;
+  final int? stepMb;
+  final String folder;
+  final int folderSizeLimitMb;
+  final Duration interval;
+  final Duration minDelayBetweenSnapshots;
+  final SnapshotCallback? onSnapshot;
```
kenzieschmoll, line 59: add dart doc to each of these members. Some of the content in the class dart doc should move here to document each value individually

more_examples/autosnapshotting/lib/main.dart
```diff
@@ -0,0 +1,134 @@
+import 'dart:io';
```
kenzieschmoll, line 1: copyright

more_examples/autosnapshotting/integration_test/app_test.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:autosnapshotting/main.dart' as app;
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+// Run:
+// flutter test integration_test/app_test.dart -d macos
+
+void main() {
```
kenzieschmoll, line 8: should this test be part of the main leak_tracker/test or part of an example app? It seems like this belongs in testing for the package itself rather than testing for an example app

more_examples/autosnapshotting/integration_test/app_test.dart
```diff
@@ -0,0 +1,38 @@
...
+  testWidgets('Snapshots are taken after reaching limit', (tester) async {
+    app.main();
+    await tester.pumpAndSettle();
+
+    final pageState =
+        tester.state<app.MyHomePageState>(find.byType(app.MyHomePage));
+    final theButton = find.byTooltip('Allocate more memory');
+
+    // Take first snapshot
+    const firstThreshold = app.memoryThresholdMb * 1024 * 1024;
```
kenzieschmoll, line 20: can you use the mbToBites extension method here (you'll have to mark the extension visible for testing)

more_examples/autosnapshotting/integration_test/app_test.dart
```diff
@@ -0,0 +1,38 @@
...
+
+    //Take second threshold
+    final secondThreshold = pageState.lastRss + app.memoryStepMb * 1024 * 1024;
+    final snapshotsLength = pageState.snapshots.length;
+    while (pageState.lastRss <= secondThreshold) {
+      await tester.tap(theButton);
+      await tester.pumpAndSettle();
+    }
+    await tester.runAsync(() => Future.delayed(const Duration(seconds: 5)));
+    expect(pageState.snapshots.length, greaterThan(snapshotsLength));
```
kenzieschmoll, line 36: why can't we assert an exact value here? (pageState.snapshots.length + 1)

more_examples/autosnapshotting/lib/main.dart
```diff
@@ -0,0 +1,134 @@
...
+      minDelayBetweenSnapshots: minDelayBetweenSnapshots,
+    );
+
+    _initConfigInfo(config);
+    autoSnapshotOnMemoryOveruse(config: config);
+  }
+
+  void _initConfigInfo(AutoSnapshottingConfig config) {
+    final stepsMb = config.stepMb;
+    _configInfo = 'thresholdMb: ${_formatter.format(config.thresholdMb)}\n'
```
kenzieschmoll, line 75: add a toString() method on `AutoSnapshottingConfig` instead and use that here

more_examples/autosnapshotting/integration_test/app_test.dart
polina-c: It is too minor to share. I do not want library code to expose utils for test code.

more_examples/autosnapshotting/integration_test/app_test.dart
polina-c: Yes. 
However, there are technical issues to move it. The test uses the folder `macos` in the root of the example.
I prefer to keep technical simplicity.

more_examples/autosnapshotting/integration_test/app_test.dart
polina-c: Visible for testing is a Flutter thing and the library is Dart.
And, I do not want leak tracker to expose this API publicly.

lib/src/autosnapshotting/autosnapshotting.dart
polina-c: Tried. It returns much smaller size. Most likely size of the folder record.

.flutter-plugins
```diff
@@ -0,0 +1,2 @@
+# This is a generated file; do not edit or check into version control.
```
kenzieschmoll, line 1: should this be added to .gitignore? same with flutter-plugins-dependencies according to comments in both those files.

doc/AUTOSNAPSHOT.md
```diff
@@ -22,9 +22,8 @@ See [example](../more_examples/autosnapshotting/).
 
 ## Limitations
 
-* **Enabled for:** Flutter
-debug and profile modes (that are Dart debug and release modes).
-* **Disabled for:** Flutter release mode (that is Dart product mode).
+* **Enabled for:** Flutter debug and profile modes (that equivalent to Dart debug and release modes).
+* **Disabled for:** Flutter release mode (that equivalent to Dart product mode).
```
kenzieschmoll, line 26: remove "that" --> "(equivalent to..."

lib/src/autosnapshotting/model.dart
```diff
@@ -20,43 +21,78 @@ class SnapshotRecord {
...
+typedef SnapshotCallback = void Function(SnapshotInfo record);
 
-/// Configures auto-snapshotting, based on the value of ProcessInfo.currentRss (dart:io).
+/// Configures auto-snapshotting, based on the value of [ProcessInfo.currentRss] (dart:io).
 ///
-/// The snapshots will be taken as soon as the value becomes more than [thresholdMb],
-/// and saved to the [folder]. The snapshots will be re-taken when the value
-/// increases more than by [stepMb], till hitting the [folderSizeLimitMb].
-/// The [folder] will be created if it does not exist.
+/// Automatic snapshots will begin to be taken when the rss value is exceeds [thresholdMb].
```
kenzieschmoll, line 28: r/value is exceeds/value exceeds

more_examples/autosnapshotting/integration_test/app_test.dart
```diff
@@ -26,13 +37,33 @@ void main() {
     expect(pageState.snapshots.length, greaterThan(0));
 
     //Take second threshold
```
kenzieschmoll, line 39: nit: space after //

more_examples/autosnapshotting/integration_test/app_test.dart
```diff
@@ -26,13 +37,33 @@ void main() {
...
+    int snapshotsLength = pageState.snapshots.length;
     while (pageState.lastRss <= secondThreshold) {
       await tester.tap(theButton);
       await tester.pumpAndSettle();
     }
     await tester.runAsync(() => Future.delayed(const Duration(seconds: 5)));
-    expect(pageState.snapshots.length, greaterThan(snapshotsLength));
+    expect(pageState.snapshots.length, greaterThan(snapshotsLength + 1));
+
+    //Check the directory limit is respected.
```
kenzieschmoll, line 49: nit: space after //

more_examples/autosnapshotting/integration_test/app_test.dart
kenzieschmoll: Where is the macos folder used? What is the technical issue that would prevent moving the test where it belongs?

kenzieschmoll: Are you referring to the call to `app.main()` that fails? When the test is under leak_tracking/integration_test/, this could an issue with how we are packaging the sample app and including it as a dependency. 

If we leave the test here, can you add a comment at the top explaining that this is not actually and example test that we expect users to duplicate, but rather a way to verify the leak_tracker behavior on a real application.

And then just want to verify one thing - is this test ran on the CI for leak_tracker?

polina-c: macos test is temporary disabled because of flutter issue: https://github.com/flutter/flutter/issues/105913
Added headless test.

About comment for the test you asked for.
The test is actually tests the example is doing what it should do - to demonstrate how auto-snapshotting works. If users modify the example, they need to modify the test.
The same way as for counter application: the autogenerated tests verifies counter is working. And, if users change the app, they should change the test.
How does it sound?

.github/workflows/macos_integration.yaml
```diff
@@ -0,0 +1,37 @@
+# TODO (polina-c): uncomment after figuring out https://github.com/flutter/flutter/issues/125636
```
kenzieschmoll, line 1: The test should not behave differently with -d flutter-tester and -d macos, so I suggest we delete this file all together.

kenzieschmoll (approved):

.github/workflows/macos_integration.yaml
polina-c: removed

PR #57, opened 2023-05-04
Fix broken link.

polina-c:


kenzieschmoll (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 4.0.3
```
kenzieschmoll, line 1: In general, I don't think it is necessary to bump the version for each commit. Is there a reason we are doing this?

polina-c: In this particular case, I want to upgrade flutter to use right link.
In general, if it is just documentation, I am skipping it.
Added test coverage for links to this PR.

PR #60, opened 2023-05-08
Make sure leak tracker does not fail in case of duplicates.

polina-c:
Contributes towards https://github.com/flutter/flutter/issues/126259

### DIFF
--- a/lib/src/shared/_primitives.dart
+++ b/lib/src/shared/_primitives.dart
@@ -5,14 +5,19 @@
 /// Result of [identityHashCode].
 typedef IdentityHashCode = int;
 
+/// Finilizer builder to mock standard [identityHashCode].
+typedef IdentityHashCoder = IdentityHashCode Function(Object object);
+
+int standardIdentityHashCoder(Object object) => identityHashCode(object);
+
 typedef ObjectGcCallback = void Function(Object code);
 
-/// Finilizer builder to mock finalizer.
+/// Finilizer builder to mock standard finalizer.
 typedef FinalizerBuilder = Finalizer<Object> Function(
   ObjectGcCallback onObjectGc,
 );
 
-Finalizer<Object> buildFinalizer(ObjectGcCallback onObjectGc) =>
+Finalizer<Object> buildStandardFinalizer(ObjectGcCallback onObjectGc) =>
     Finalizer<Object>(onObjectGc);
 
 String fullClassName({
--- a/lib/src/shared/_primitives.dart
+++ b/lib/src/shared/_primitives.dart
@@ -12,7 +12,7 @@ int standardIdentityHashCoder(Object object) => identityHashCode(object);
 
 typedef ObjectGcCallback = void Function(Object code);
 
-/// Finilizer builder to mock standard finalizer.
+/// Finilizer builder to mock standard [Finalizer].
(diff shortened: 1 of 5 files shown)


test/dart/leak_tracking/_object_tracker_test.dart
```diff
@@ -223,6 +223,16 @@ void main() {
...
+    late _MockFinalizerBuilder finalizerBuilder;
+    late _MockGcCounter gcCounter;
+    late ObjectTracker tracker;
+    const disposalTimeBuffer = Duration(milliseconds: 100);
+
+    /// Emulates GC.
+    void gc(Object object) {
+      finalizerBuilder.finalizer.finalize(identityHashCode(object));
+    }
+
```
kenzieschmoll, line 235: this is all duplicated. Why can't it be defined outside of the groups so that it can be reused by both?

polina-c: I plan to create one more group that behaves differently.
So I want these groups to be independent so that the shared things can change in different directions if needed.

kenzieschmoll: If an additional group needed to create its own variables, those could still be created inside the group and then they'd be the closest in scope for the group's tests, so we should need to duplicate everywhere when only one group needs tighter scoped-variables.

polina-c: Added actual test and fix to this PR.
Reorganized the test do decrease duplicated code.

Now only three lines duplicate (declaration for finalizerBuilder, gcCounter and tracker). This is what we normally duplicate between libraries, so it should be ok to duplicate between groups.

How does it sound?

polina-c: I prefer it to be visible to the reader what mocks are used by what tests. I want tests with the same time of mocking to be grouped. It will make maintenance easier.

lib/src/shared/_primitives.dart
```diff
@@ -5,14 +5,19 @@
...
 
+/// Finilizer builder to mock standard [identityHashCode].
+typedef IdentityHashCoder = IdentityHashCode Function(Object object);
+
+int standardIdentityHashCoder(Object object) => identityHashCode(object);
+
 typedef ObjectGcCallback = void Function(Object code);
 
-/// Finilizer builder to mock finalizer.
+/// Finilizer builder to mock standard finalizer.
```
kenzieschmoll, line 15: sp: Finalizer

test/dart/leak_tracking/_object_tracker_test.dart
```diff
@@ -223,6 +251,15 @@ void main() {
...
 
   group('$ObjectTracker with stack traces', () {
+    late _MockFinalizerBuilder finalizerBuilder;
+    late _MockGcCounter gcCounter;
+    late ObjectTracker tracker;
+
+    /// Emulates GC.
+    void gc(Object object) {
+      finalizerBuilder.finalizer.finalize(identityHashCode(object));
+    }
```
kenzieschmoll, line 261: nit: since this is duplicated, what do you think about pulling this outside of the groups for shared use, and passing the `finalizerBuilder` as a parameter?

kenzieschmoll (approved):
couple comments but lgtm

PR #66, opened 2023-05-18
blast_repo fixes

devoncarew:
dependabot

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -3,10 +3,12 @@ version: 2
 enable-beta-ecosystems: true
 
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
+  - package-ecosystem: github-actions
+    directory: /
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -3,10 +3,12 @@ version: 2
 enable-beta-ecosystems: true
 
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
+  - package-ecosystem: github-actions
+    directory: /
     schedule:
-      interval: "weekly"
+      interval: monthly
```
polina-c, line 9: Why not daily?

polina-c (approved):

.github/dependabot.yaml
devoncarew: This keeps us up to date but reduces the number of PRs we receive (we do get any security related updates out of band w/ this frequency).

PR #97, opened 2023-07-19
Add details about hidden leaks.

polina-c:


doc/TROUBLESHOOT.md
```diff
@@ -101,6 +101,13 @@ Otherwise, the reference to the first non-needed object on the path (`staticX`,
 staticX -> A -> B -> disposedD
 ```
 
+Be careful not to fall into pattern to limit your fix to just releasing the reference to the disposed object. 
+Instead, you want to release most left that is not needed any more (like `staticX`, `A` or `B`).
```
gspencergoog, line 105: I'm not sure what you mean be "most left" here.  Do you mean that you want to release as much as possible that remains, or do you mean you want to release the leftmost (most specific) items?

doc/TROUBLESHOOT.md
```diff
@@ -101,6 +101,35 @@ Otherwise, the reference to the first non-needed object on the path (`staticX`,
 staticX -> A -> B -> disposedD
 ```
 
+Be careful not to fall into pattern to limit your fix to just releasing the reference to the disposed object.
```
dnfield, line 104: Consider telling the user what they should do instead of what they should not do.

e.g. `The leak tracker does not detect all leaks, only the leaks of objects it explicitly is tracking. When fixing detected leaks, make sure to look for other objects that may have been leaked without being visible to the tracker.`

An code example with a `BAD` and `GOOD` example would probably help here too.

doc/TROUBLESHOOT.md
```diff
@@ -101,6 +101,35 @@ Otherwise, the reference to the first non-needed object on the path (`staticX`,
 staticX -> A -> B -> disposedD
 ```
 
+Be careful not to fall into pattern to limit your fix to just releasing the reference to the disposed object. 
+Instead, you need to release just most left object that is not needed any more (first check `staticX`, then `A`, then `B`). All objects
```
dnfield, line 105: I'm not sure what a `left most object` is. Maybe the root object? Root might not be right either - what happens when there is a cycle for example?

doc/TROUBLESHOOT.md
```diff
@@ -101,6 +101,35 @@ Otherwise, the reference to the first non-needed object on the path (`staticX`,
...
+referenced from it will also become unreachable, and thus available for garbage collection.
+
+Remember, not all leaks are detected, but for 
+[the leak tracked objects]([staticX -> A -> B](https://github.com/dart-lang/leak_tracker/blob/main/doc/DETECT.md#by-tracked-classes)) 
+only. So, if you just release the 
+reference to disposed object, you will fix the detected leak and hide the undetected ones.
+
+And, there is no point to release reference to child inside the method `dispose` of the parent, because
+not needed parent should be released together with its disposal. If
+your fix for a leak is like this, you are defenitely hiding the leak:
```
dnfield: I'm not sure this is true.

If the child is not nulled out, it may not be eligible for garbage collection (even if it has released some resource in dispose). And if it is _not_ nulled out, other code in the class that depends on its nullability to determine whether it can be used or not may fail.

For example, it's pretty common in Flutter code to mark some native object as nullable, and once it is disposed null it out so it doesn't get used again.

polina-c: I am not saying it is wrong to null out. I am saying it is wrong to null out in `dispose`.
If parent is disposed, it should become non-used, unreachable and thus to be garbage collected together with child.

Another side effect of nulling in dispose is that you have to make the field nullable, that adds unnecessary complexity to the code and eliminates non-nullable optimizations.

How does it sound?

doc/TROUBLESHOOT.md
polina-c: There are no cycles in retaining path and they are always from root. Updated the text.

doc/TROUBLESHOOT.md
```diff
@@ -98,9 +98,37 @@ If `A` and `B` are still needed, `B` should assign null to the variable that ref
...
 
+The leak tracker does not detect all leaks, only the leaks of objects
+[it explicitly is tracking](https://github.com/dart-lang/leak_tracker/blob/main/doc/DETECT.md#by-tracked-classes).
+When fixing detected leaks, make sure to look for other objects that may have been leaked without being visible to the tracker.
+
+To fix all leaks, you need to release closest to the root object on retaining path
+(first check `staticX`, then `A`, then `B`), that is not needed any more. All objects referenced from it will
+also become unreachable, and thus available for garbage collection.
+
+One of signes that some leakes are still, is fixing a leak by releasing link to child in the parent's `dispose`,
```
gspencergoog, line 112:
```suggestion
One of signs that some leaks still exist is fixing a leak by releasing link to child in the parent's `dispose`,
```

doc/TROUBLESHOOT.md
```diff
@@ -98,9 +98,37 @@ If `A` and `B` are still needed, `B` should assign null to the variable that ref
...
+because link to not needed parent should be released itself, together with its disposal. If
```
gspencergoog, line 113:
```suggestion
because the link to a parent that is no longer needed should be released when it is disposed. If
```

doc/TROUBLESHOOT.md
polina-c: I see. There are pros and cons nulling out member in dispose.

However, the statement here is: a leak fix should not be nulling out in `dispose`, because, while it makes the leak smaller, it does not fix it.

This means it is ok to null out the member in dispose in a separate PR, but not in order to fox the leak. Am I missing something?

dnfield: Yeah, I think pointing out that nulling it out isn't the right way to fix the leak, but I would avoid calling that style bad - it's used frequently in the flutter framework for the reasons I mentioned

polina-c: Modified. Is it better now?

dnfield (approved):

PR #98, opened 2023-07-19
Improve performance.

polina-c:
1. Enabled customization for gcCountBuffer, so that tests can use value 1 instead of 3.
2. Fine tuned a way to allocate memory in forceGC, to make it working faster
3. Test covered performance of forceGC

Also removed non published versions from changelog.

### DIFF
--- a/pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
+++ b/pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
@@ -78,6 +78,15 @@ class LeakDiagnosticConfig {
       classesToCollectStackTraceOnDisposal.contains(classname);
 }
 
+/// The default value for number of full GC cycles, enough for a non reachable object to be GCed.
+///
+/// It is pessimistic assuming that user will want to
+/// detect leaks not more often than a second.
+///
+/// Theoretically, 2 should be enough, however it gives false positives
+/// if there is no activity in the application for ~5 minutes.
+const defaultGcCountBuffer = 3;
+
 class LeakTrackingConfiguration {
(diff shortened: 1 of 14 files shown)


pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
```diff
@@ -78,6 +78,15 @@ class LeakDiagnosticConfig {
...
 }
 
+/// The default value for number of full GC cycles, enough for a non reachable object to be GCed.
+///
+/// It is pessimistic assuming that user will want to
+/// detect leaks not more often than a second.
+///
+/// Theoretically, 2 should be enough, however it gives false positives
+/// if there is no activity in the application for ~5 minutes.
+const defaultGcCountBuffer = 3;
```
devoncarew, line 88: I'm not 100% if this is public API or not (whether it's exported from somewhere), but if it is we should mention this in the changelog.

pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
```diff
@@ -86,6 +95,7 @@ class LeakTrackingConfiguration {
     this.checkPeriod = const Duration(seconds: 1),
     this.disposalTimeBuffer = const Duration(milliseconds: 100),
     this.leakDiagnosticConfig = const LeakDiagnosticConfig(),
+    this.gcCountBuffer = defaultGcCountBuffer,
```
devoncarew, line 98: Same point here (if `LeakTrackingConfiguration` is public we should mention the addition of a new ctor param).

devoncarew (approved):

PR #99, opened 2023-07-20
Fix and test cover case of customized `gcCountBuffer`

polina-c:


CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+# Contributing code
```
devoncarew, line 1: This will override the dart-lang/ default at https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md (the UI will show this file instead of that one).

I would ether reference the fuller doc from here, or, remove this file and in-line the bit about enabling logs in the readme.

pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
```diff
@@ -95,7 +95,7 @@ class LeakTrackingConfiguration {
     this.checkPeriod = const Duration(seconds: 1),
     this.disposalTimeBuffer = const Duration(milliseconds: 100),
     this.leakDiagnosticConfig = const LeakDiagnosticConfig(),
-    this.gcCountBuffer = defaultGcCountBuffer,
+    required this.gcCountBuffer,
```
devoncarew, line 98: We should mention in the changelog that this is now required.

pkgs/leak_tracker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 8.0.3
+
+* Fix and test cover case of customized `gcCountBuffer`.
```
devoncarew, line 3: "Fix an issue with custom `gcCountBuffer` values."?

pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
devoncarew: (and / or ensure that the changes mentioned there cover user-facing ones, like api changes)

devoncarew (commented):
A few comments related to the changelog and over-ridding the main dart-lang/ contributing.md file (note also the test failures).

pkgs/leak_tracker/lib/src/leak_tracking/leak_tracker_model.dart
polina-c: reverted

devoncarew (approved):
👍

PR #100, opened 2023-07-21
Improve doc comments.

polina-c:


pkgs/leak_tracker/lib/src/leak_tracking/orchestration.dart
```diff
@@ -45,8 +45,13 @@ class MemoryLeaksDetectedError extends StateError {
 /// to wait infinitely for the forced garbage collection, that is needed
 /// to analyse results.
 ///
-/// [gcCountBuffer] is number of full GC cycles, enough for a non reachable object to be GCed.
-///
+/// [gcCountBuffer] is number of full GC cycles, that should be enough for
```
goderbauer, line 48: Thanks for the explanation! I guess, I am still confused by what the "Buffer" in the name is supposed to mean. Would this be more appropriately named "gcCycleCount" of "numberOfGcCycles"?

goderbauer (approved):
LGTM

pkgs/leak_tracker/lib/src/leak_tracking/orchestration.dart
polina-c: good point. Will rename. It will be breaking change, but it is ok while leak tracker is work in progress.

PR #115, opened 2023-08-04
Refactor dispatcher.

polina-c:
Goals of refactoring:
1. Make dispatcher to stop being concerned about ObjectTracker
2. Make all calls to dispatchObjectEvent going through dispatchObjectCreated and dispatchObjectCreated in LeakTracking

No functional changes.


pkgs/leak_tracker_flutter_test/test/test_infra/event_tracker.dart
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```
CoderDake, line 1: does this need to be 2023?

CoderDake (approved):

pkgs/leak_tracker_flutter_test/test/test_infra/event_tracker.dart
polina-c: It is renamed file. So, it's ok.

PR #117, opened 2023-08-07
Stop failing if object is disposed twice.

polina-c:
The failures affect running application. 

This fix will unblock the PR: https://github.com/flutter/flutter/pull/131998

### DIFF
--- a/pkgs/leak_tracker/CHANGELOG.md
+++ b/pkgs/leak_tracker/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.3
+
+* Stop failing if an object is disposed twice.
+
 # 9.0.2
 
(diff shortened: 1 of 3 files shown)


kenzieschmoll:
Instead of allowing objects to be disposed twice, shouldn't we prevent them from being disposed twice?

polina-c:
> Instead of allowing objects to be disposed twice, shouldn't we prevent them from being disposed twice?

It may be good idea, but for future scope of leak_tracker. Right now it is not leak tracker responsibility. 
But, if we decide to add it to scope, failure is not right way to handle double disposal, because leak_tracker should not change the application or test behaviour.  

So, if we want to handle it, it requires additional discussion and design.
Right now i want to turn off failures to stop blocking bug PR that may get merge conflicts.

PR #122, opened 2023-08-13
Fix issue of using wrong settings for a phase.

polina-c:
Instead of using settings for current phase, ObjectTracker should use settings for a phase associated with the tracked object.

### DIFF
--- a/pkgs/leak_tracker/CHANGELOG.md
+++ b/pkgs/leak_tracker/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.5
+
+* Fix issue of using wrong settings for a phase.
+
 # 9.0.4
 
 * Make it possible to disable tracking for a type of leak.
--- a/pkgs/leak_tracker/CHANGELOG.md
+++ b/pkgs/leak_tracker/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 9.0.5
 
-* Fix issue of using wrong settings for a phase.
+* Fix issue of using wrong settings for a phase, so that the tracker uses settings
+at time of object tracking start, instead of current configuration.
 
 # 9.0.4
 
--- a/pkgs/leak_tracker/lib/src/shared/_util.dart
+++ b/pkgs/leak_tracker/lib/src/shared/_util.dart
@@ -32,6 +32,13 @@ T cast<T>(value) {
   );
 }
 
+extension IterableExtensions<T> on Iterable<T> {
+  T? get onlyOrNull {
+    if (length > 1) throw StateError('Lenth should not be more than one.');
+    return firstOrNull;
+  }
+}
+
 void printToConsole(Object message) {
   // ignore: avoid_print, dart:io is not available in web
   print('leak_tracker: $message');
--- a/pkgs/leak_tracker/lib/src/shared/_util.dart
+++ b/pkgs/leak_tracker/lib/src/shared/_util.dart
@@ -33,6 +33,9 @@ T cast<T>(value) {
 }
 
 extension IterableExtensions<T> on Iterable<T> {
+  /// Returns the item or null, assuming that the length of the itrable os 0 or 1.
+  // The name is consistent woth other methods names on iterable like
+  // `firstOrNull, lastOrNull and singleOrNull`.
   T? get onlyOrNull {
     if (length > 1) throw StateError('Lenth should not be more than one.');
     return firstOrNull;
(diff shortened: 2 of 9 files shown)


pkgs/leak_tracker/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 9.0.5
+
+* Fix issue of using wrong settings for a phase.
```
CoderDake, line 3: A little more context could be helpful here

pkgs/leak_tracker/lib/src/shared/_util.dart
```diff
@@ -32,6 +32,13 @@ T cast<T>(value) {
   );
 }
 
+extension IterableExtensions<T> on Iterable<T> {
+  T? get onlyOrNull {
+    if (length > 1) throw StateError('Lenth should not be more than one.');
```
CoderDake, line 37:
```suggestion
    if (length > 1) throw StateError('Length should not be more than one.');
```

pkgs/leak_tracker/lib/src/shared/_util.dart
```diff
@@ -32,6 +32,13 @@ T cast<T>(value) {
   );
 }
 
+extension IterableExtensions<T> on Iterable<T> {
+  T? get onlyOrNull {
```
CoderDake, line 36: This could use a doc. The name might need some work as well. It isn't obvious at first glance what this tries to accomplish.

pkgs/leak_tracker/test/tests/leak_tracking/_object_tracker_test.dart
```diff
@@ -375,96 +389,197 @@ void main() {
...
-      allObjects.forEach(finalizerBuilder.gc);
-
-      // Check leaks are collected only for leaking object that
-      // was registered in the phase.
-      final leaks = await tracker.collectLeaks();
-      expect(leaks.total, 1);
-      final theLeak = leaks.byType[LeakType.notDisposed]!.single;
-      expect(theLeak.code, identityHashCode(objectInPhaseLeaking));
-    });
+    void checkContext(
```
CoderDake, line 448: to help readability of tests, it might be better to make these helper functions as custom matchers

pkgs/leak_tracker/test/tests/leak_tracking/_object_tracker_test.dart
```diff
@@ -375,96 +389,197 @@ void main() {
...
+        key: ContextKeys.retainingPath,
+      );
+    }
+
+    for (var gced in [
+      true,
+      //false,
+    ]) {
+      for (var disposed in [
+        //true,
```
CoderDake, line 524:
```suggestion
      false,
    ]) {
      for (var disposed in [
        true,
```
Should these be uncommented?

pkgs/leak_tracker/lib/src/shared/_util.dart
polina-c: Added doc. The name is consistent with other names on iterable:

Screenshot 2023-08-14 at 12 22 59 PM

pkgs/leak_tracker/test/tests/leak_tracking/_object_tracker_test.dart
polina-c: Interesting. I can think about it and maybe implement in other PRs.
Did you ever create custom matchers, not for publishing, but for local use, in your tests? If yes, can you give example?

pkgs/leak_tracker/test/tests/leak_tracking/_object_tracker_test.dart
polina-c: yes!

pkgs/leak_tracker/test/tests/leak_tracking/_object_tracker_test.dart
polina-c: I have feeling that custom matcher implementation is more complicated than helper method, and create overhead for debugging the matcher itself, so it would be overdesign to do just for local use.

CoderDake: I haven't written any myself yet, but it looks like it is easy enough: https://adevait.com/flutter/write-flutter-tests-custom-matchers

A couple years back I did a bunch of reading into unit test best practices, they mention that using helpers should be avoided since it makes tests harder to read. Something about readability and keeping the logic inside the main context.

All in all if these tests get the job done, then that is just a bit of a nit pick on my part

CoderDake (approved):

pkgs/leak_tracker/test/tests/leak_tracking/_object_tracker_test.dart
polina-c: Is this practice recommended by dart team? The article you linked is not.
The problem with matchers is that they are not strongly typed, so it is extra effort to do type promotion.

PR #128, opened 2023-08-15
Enable connection to vm service for flutter testing and define matcher for reporting for memory allocations.

polina-c:


pkgs/leak_tracker/lib/src/leak_tracking/_primitives/_retaining_path/_connection.dart
```diff
@@ -18,26 +18,30 @@ class Connection {
...
   }
 
+  return uri;
+}
+
+/// Connects to vm service protocol.
+///
+/// If it is not found, tries to start it.
+Future<Connection> connect() async {
+  _log.info('Connecting to vm service protocol...');
```
bkonyi, line 41: Nit: VM

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:leak_tracker_testing/leak_tracker_testing.dart';
+
+/// Checks if the object dispatches events to `MemoryAllocations.instance`.
+const Matcher dispatchesMemoryEvents = _DispatchesMemoryEvents();
+
+class _DispatchesMemoryEvents extends Matcher {
+  const _DispatchesMemoryEvents();
+
+  static const _key = 'description';
```
bkonyi, line 15: Is this a placeholder?

pkgs/leak_tracker_testing/lib/src/model.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Object lifecycle.
+///
+/// Is used to test if an object dispatched deeded events at creation and disposal.
+class ObjectLyfecycle {
```
bkonyi, line 8: Should this be `ObjectLifecycle`?

pkgs/leak_tracker_testing/lib/src/model.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Object lifecycle.
+///
+/// Is used to test if an object dispatched deeded events at creation and disposal.
```
bkonyi, line 7: Nit: I'm not sure what "dispatched deeded events" is. Is "deeded" necessary?

pkgs/leak_tracker_flutter_testing/pubspec.yaml
```diff
@@ -12,11 +12,12 @@ dependencies:
   flutter_test:
     sdk: flutter
   leak_tracker:  '>=9.0.4 <10.0.0'
-  leak_tracker_testing: '>=1.0.0 <2.0.0'
+  leak_tracker_testing: '>=1.0.4 <2.0.0'
```
bkonyi, line 15: Ubernit: can this just be `^1.0.4`? Typically we avoid explicit ranges that only span a single major revision.

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:leak_tracker_testing/leak_tracker_testing.dart';
+
+/// Checks if the object dispatches events to `MemoryAllocations.instance`.
+const Matcher dispatchesMemoryEvents = _DispatchesMemoryEvents();
```
bkonyi, line 10: Most matchers that don't take arguments use the verbiage `is$Something`. Would something like `isDisposed` work here?

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
polina-c: It is not. I just needed some key.

pkgs/leak_tracker_testing/lib/src/model.dart
polina-c: I little redesigned the matcher and the type is not needed any more.

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
polina-c: After some thinking redesigned to have arguments. Seems to be cleaner.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/_retaining_path/_connection.dart
```diff
@@ -18,26 +18,30 @@ class Connection {
...
+      'Could not start VM service. If you are running `flutter test`, pass the flag `--enable-vmservice`',
     );
   }
 
+  return uri;
+}
+
+/// Connects to vm service protocol.
+///
+/// If it is not found, tries to start it.
```
CoderDake, line 39:
```suggestion
/// If the VM service is not found, tries to start it.
```

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/_retaining_path/_connection.dart
```diff
@@ -18,26 +18,30 @@ class Connection {
...
     throw StateError(
-      'Leak troubleshooting is not available. Run your application with flag "--debug" to enable VM service.',
+      'Could not start VM service. If you are running `flutter test`, pass the flag `--enable-vmservice`',
     );
   }
 
+  return uri;
+}
+
+/// Connects to vm service protocol.
```
CoderDake, line 37:
```suggestion
/// Connects to VM service protocol.
```

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
```diff
@@ -0,0 +1,67 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+/// Checks if the object dispatches events to `MemoryAllocations.instance`.
+///
+/// [createAndDispose] should create and dispose the object.
+Matcher dispatchesMemoryEvents(Function() createAndDispose) =>
```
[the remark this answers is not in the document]
CoderDake: I think I like the idea of Matchers instead of helpers even though it might seem odd at first. Helpers obscure important aspects of the test that provide context as someone is reading through a test.
On the other hand matchers provide verbosity and readability, while reading a test, even though some of the work is hidden behind it. 

In this specific case it does feel like [checking if this matcher throws an error](https://github.com/dart-lang/leak_tracker/pull/128/files#diff-def4ff981195fc4483ed7d9a3cc2f6acadd9ac7346fb5a2b5be35a9612f70fd1R37) is a bit of a misuse of the matcher pattern. I wonder if creating a `failsToDispatchMemoryEvents` matcher would be a better way to make that behaviour accessible.

I think precident for using matchers like this has been set already, for example the matcher packages makes behaviour similar to these memory matchers with:
- [emits](https://pub.dev/documentation/matcher/latest/expect/emits.html)
- [emitsError](https://pub.dev/documentation/matcher/latest/expect/emitsError.html)

These emit matchers can be seen in the [stream matchers section](https://pub.dev/packages/matcher#stream-matchers)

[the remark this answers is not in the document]
bkonyi: I'm not saying we shouldn't have a `Matcher` for this, but `Matcher`s should be used to verify results, not actually perform the operations that create the results. In this particular case, the `actual` parameter of `expect(...)` is not the result of an operation but a hard-coded type. The matcher only serves to verify that `actual is Type == true` but then proceeds to do nothing with it before actually running code that should be used to generate `actual` in the first place.

This block of code shouldn't be inside a matcher since it's generating a result rather than verifying one:

```dart
MemoryAllocations.instance.addListener(listener);
createAndDispose();
MemoryAllocations.instance.removeListener(listener);
```

Wrapping that code in a helper that generates a result that can be verified against a `Matcher` isn't any less readable than hiding it in the matcher in the first place.

> In this specific case it does feel like checking if this matcher throws an error is a bit of a misuse of the matcher pattern. I wonder if creating a failsToDispatchMemoryEvents matcher would be a better way to make that behaviour accessible

This can be achieved by wrapping a matcher with `isNot` (e.g., `expect('String', isNot(equal('Foo')))`).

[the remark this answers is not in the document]
CoderDake: hmm now that i'm reading a bit more into this I wonder if I've missunderstood who would be using this. I'm going to loop back on this in an hour after lunch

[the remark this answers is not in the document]
CoderDake: Hmm what if the matcher was built like the [`throwA`](https://pub.dev/packages/matcher#exception-matchers) Matcher. The function could be on the left side of the expect, then we could still do all of the necessary setup and teardown to do the test.

Using the matcher would look a lot more like you expect it:
```dart
expect(
  () => _TrackedClass().dispose(),
  dispatchesMemoryEventsFor(_TrackedClass),
)
expect(
  () => {},
  isNot(dispatchesMemoryEventsFor(_TrackedClass)),
)
```

The Matcher would look something like this instead:
```dart
class _DispatchesMemoryEvents extends Matcher {
  const _DispatchesMemoryEvents(this.type);

  final Type type;
  static const _key = 'description';

  @override
  bool matches(Function block, Map matchState) {
    final events = <ObjectEvent>[];

    void listener(ObjectEvent event) {
      if (event.object.runtimeType == type) {
        events.add(event);
      }
    }

    MemoryAllocations.instance.addListener(listener);
    block();
    MemoryAllocations.instance.removeListener(listener);

    if (events.length == 2 &&
        events.first is ObjectCreated &&
        events.last is ObjectDisposed) {
      return true;
    }

    matchState[_key] =
        'createAndDispose is expected to dispatch two events to $MemoryAllocations.instance,'
        ' for the type $type,'
        ' first $ObjectCreated and then $ObjectDisposed.\n'
        'Actually it dispatched ${events.length} events:\n$events';

    return false;
  }

...
}
```

polina-c:
> hmm now that i'm reading a bit more into this I wonder if I've missunderstood who would be using this. I'm going to loop
> back on this in an hour after lunch

Added comment.

polina-c:
Flipped type and function. Now it is like `throwsA`. Is it better?

pkgs/leak_tracker_flutter_testing/test/tests/matchers_test.dart
```diff
@@ -23,17 +23,14 @@ class _TrackedClass {
 void main() {
   test('dispatchesMemoryEvents success', () {
     expect(
-      _TrackedClass,
-      dispatchesMemoryEvents(() => _TrackedClass().dispose()),
+      () => _TrackedClass().dispose(),
```
bkonyi, line 26: I definitely like this better!

bkonyi (approved):

PR #129, opened 2023-08-16
Add debugging construcstors to LeakTrackingTestConfig, per leak type.

polina-c:


pkgs/leak_tracker_flutter_testing/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.0.2
+
+* Add debugging construcstors to LeakTrackingTestConfig, per leak type.
```
justinmc, line 3: "construcstors" => "constructors"

pkgs/leak_tracker_flutter_testing/lib/src/model.dart
```diff
@@ -57,7 +57,36 @@ class LeakTrackingTestConfig {
...
+    this.notGCedAllowList = const <String, int>{},
+    this.notDisposedAllowList = const <String, int>{},
+    this.allowAllNotDisposed = false,
+    this.allowAllNotGCed = false,
+  });
+
+  /// Creates a new instance for debugging notDisposed leaks.
+  ///
+  /// This configuration will collect stack traces on start and disposal,
+  /// and retaining path for notGCed objects.
```
justinmc, line 78: "retaining" => "retain"

justinmc (approved):
LGTM :+1:  I'm no expert on the leak tracker but looks fine at a high level!

pkgs/leak_tracker_flutter_testing/lib/src/model.dart
polina-c: `retaining path` is a concept in memory management: https://docs.flutter.dev/tools/devtools/memory#root-object-retaining-path-and-reachability

replaced with 'the retaining path' to show it is noun.

justinmc: Ah got it, thanks!

CoderDake (approved):

PR #140, opened 2023-09-01
Update matcher to handle async callbacks.

polina-c:


pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
```diff
@@ -5,53 +5,50 @@
...
-/// The memory events are used by tools like leak_tracker for diagnostics.
-///
-/// The matcher checks that the object object is instrumented properly,
-/// dispatches two events to `MemoryAllocations.instance`,
-/// first `ObjectCreated` and then `ObjectDisposed`.
-Matcher dispatchesMemoryEvents(Type type) {
-  return _DispatchesMemoryEvents(type);
+/// Invokes [callback] and collects events dispatched to [MemoryAllocations.instance] for [type].
+Future<List<ObjectEvent>> memoryEvents(
+  Future<void> Function() callback,
```
bkonyi, line 10: This should be `FutureOr<void> Function() callback`

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
```diff
@@ -5,53 +5,50 @@
...
+  await callback();
+  MemoryAllocations.instance.removeListener(listener);
+
+  return events;
 }
 
-class _DispatchesMemoryEvents extends Matcher {
-  const _DispatchesMemoryEvents(this.type);
+/// Checks if Iterable<ObjectEvent> contains two events, first `ObjectCreated` and then `ObjectDisposed`.
+Matcher areCreateAndDispose = const _AreCreateAndDispose();
```
bkonyi, line 29: Nit: consider naming this `isCreatedAndDisposed` and the actual `Matcher` `_IsCreatedAndDisposed`.

pkgs/leak_tracker_flutter_testing/test/tests/matchers_test.dart
```diff
@@ -21,17 +21,21 @@ class _TrackedClass {
 }
 
 void main() {
-  test('dispatchesMemoryEvents success', () {
-    expect(
-      () => _TrackedClass().dispose(),
-      dispatchesMemoryEvents(_TrackedClass),
+  test('dispatchesMemoryEvents success', () async {
+    await expectLater(
```
bkonyi, line 25: I don't think you need to await `memoryEvents` below as `expectLater` will do that.

pkgs/leak_tracker_flutter_testing/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.0.0
```
bkonyi, line 1: Can we just retract 1.0.3 and re-release with this, or has it already been picked up externally?

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
polina-c: The validated item is changed now. Now it is list of events. And matcher verifies that events 'are create and dispose'.

pkgs/leak_tracker_flutter_testing/test/tests/matchers_test.dart
polina-c: If I do not await, the matcher gets Future<List<ObjectEvent>>.

pkgs/leak_tracker_flutter_testing/CHANGELOG.md
polina-c: Let's try.

It is in use in Flutter Framework. Yes, I will need to handle breaking changes anyway.
I am little afraid to get some side effects of version resolution.

PR #145, opened 2023-09-19
Remove failed assert.

polina-c:
Failure: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20framework_tests_libraries/46558/overview

### DIFF
--- a/pkgs/leak_tracker/CHANGELOG.md
+++ b/pkgs/leak_tracker/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.7
+
+* Remove failed assert.
+
 # 9.0.6
 
 * Improve error reporting for connection to vm service.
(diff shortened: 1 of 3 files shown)


jonahwilliams (changes requested):
It doesn't seem like you should be removing an assert that is checking that leak tracker is working?

polina-c:
> It doesn't seem like you should be removing an assert that is checking that leak tracker is working?

This assert should never fail because of tracked objects. If it failed, this means there is algorithmic error. And it seems it failed because of duplicated hash code. Leak tracker expects duplicated, but it seems not very reliably. And this story is not easy to test.

I plan to refactor leak_tracker to stop counting on hashes. But for now I do not want this noise in flutter test failures. There is enough flakiness in flutter itself :)

Makes sense?

jonahwilliams (commented):
I guess, but seems like a pretty severe issue. if its not happening frequently, you might just want to leave it as is

polina-c:
> I guess, but seems like a pretty severe issue. if its not happening frequently, you might just want to leave it as is

Maybe. Lets have this PR ready to merge. I will do some adjustments and maybe we will not need it.
Converted to draft.

polina-c:
I updated leak_tracker to stop identifying objects by hash codes. So, this PR should not be needed.

PR #146, opened 2023-09-20
Start using ObjectRecord instead of identityHashCode to identify objects.

polina-c:


pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
```diff
@@ -70,38 +65,23 @@ class ObjectTracker implements LeakProvider {
...
 
-    _objects.notGCed[code] = record;
+    _finalizer.attach(object, record);
 
-    _objects.assertRecordIntegrity(code);
+    _objects.assertRecordIntegrity(record);
   }
 
-  void _onOobjectGarbageCollected(Object code) {
+  void _onOobjectGarbageCollected(Object record) {
```
derekxu16, line 80: There's an extra 'o', i.e. `Oobject` -> `Object`.

derekxu16 (approved):
LGTM after my one comment is addressed.

PR #149, opened 2023-09-26
Fix flakiness in detection of notDisposed leaks in tests. 

polina-c:
Contributes to https://github.com/flutter/flutter/issues/135716

This is follow up to fix leak tracker flakiness: https://github.com/flutter/flutter/pull/135394

This PR resolves false negatives, so it is breaking change.
Objects, not disposed by the end of the test, should be marked as leaks, even if they are not GCed yet.

The upgrade to this version should be handled manually, because it may result in some tests failing.

### DIFF
--- a/pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
+++ b/pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
@@ -107,7 +107,7 @@ class ObjectTracker implements LeakProvider {
   /// Is used to make sure all disposables are disposed by the the end of the test.
   void declareAllNotDisposedAsLeaks() {
     throwIfDisposed();
-    // We need this temporary storage to avoid errror 'concurrent modification during iteration'
+    // We need this temporary storage to avoid error 'concurrent modification during iteration'
     // for internal iterables in `_objects.notGCed`.
     final notGCedAndNotDisposed = <ObjectRecord>[];
     _objects.notGCed.forEach((record) {
--- a/pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
+++ b/pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
@@ -62,6 +62,7 @@ class ObjectTracker implements LeakProvider {
     required String trackedClass,
     required PhaseSettings phase,
   }) {
+    //print('!!!!! started tracking ${identityHashCode(object)}');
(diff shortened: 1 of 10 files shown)


pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
```diff
@@ -98,6 +102,18 @@ class ObjectTracker implements LeakProvider {
     _objects.assertRecordIntegrity(record);
   }
 
+  /// Declares all not disposed objects as leaks, even if they are not GCed yet.
+  ///
+  /// Is used to make sure all disposables are disposed by the the end of the test.
+  void declareAllNotDisposedAsLeaks() {
+    throwIfDisposed();
+    final notGCedAndNotDisposed = <ObjectRecord>[];
+    _objects.notGCed.forEach((record) {
```
bkonyi, line 111: Nit: lots of projects have lints recommending that we don't use `forEach`. Even if we don't have it enabled for this project, consider doing this instead:

```dart
for (final record in _objects.notGCed) {
    if (!record.isDisposed) {
      _declareNotDisposedLeak(record);
    }
}
```

bkonyi (approved):

pkgs/leak_tracker/lib/src/leak_tracking/_object_tracker.dart
polina-c: The lint is enabled for this project. notGCed is not iterable though. I will check if it is easy to make it iterable.

bkonyi: Oh interesting. At the very least, I don't think you need `noGCedAndNotDisposed`, but you could instead do:

```dart
_objects.notGCed.forEach((record) {
  if (!record.isDisposed) {
    _declareNotDisposedLeak(record);
  }
});
```

polina-c: Yes, looks as optimizable. Added comment why I did it:

```
// We need this temporary storage to avoid errror 'concurrent modification during iteration'
// for internal iterables in `_objects.notGCed`.
```

polina-c: And I checked: Iterable wants 27 methods to be implemented. Seems to be too heavy.

CoderDake (approved):

PR #152, opened 2023-10-10
Create adjustable test settings.

polina-c:
Contributes to https://github.com/flutter/flutter/issues/135856

Draft of PR for Flutter to consume the changes: https://github.com/flutter/flutter/pull/136692.

After merging this PR, I will update flutter/test, to have 8% of tests that are not leak free yet looking like this: 

```
    testWidgets('initialTimerDuration falls within limit',
    experimentalLeakTracking: LeakTrackingForTests.ignored(),
    (WidgetTester tester) async {
        ...
```


pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+  ///
+  /// Items in [classes] will be removed from all skip lists.
+  static LeakTrackingForTestsSettings withTracked({
+    List<String> notGCed = const [],
+    List<String> notDisposed = const [],
+    List<String> classes = const [],
+  }) {
+    final result = settings.copyWith(
+      leakSkipLists: LeakSkipLists(
+        notGCed: settings.leakSkipLists.notGCed.track(notGCed).track(classes),
```
kenzieschmoll, line 108: how about `.track([...notGCed, ...classes])` so we remove a call to `track`? same below.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+          ),
+        ),
+      ),
+    );
+  }
+
+  /// Returns copy of [settings] with reduced skip lists.
+  ///
+  /// Items in [classes] will be removed from all skip lists.
+  static LeakTrackingForTestsSettings withTracked({
```
kenzieschmoll, line 101: naming: `copyWithTrackedLeaks` to be consistent with name suggestion above

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+            .track(classes),
+      ),
+    );
+    return result;
+  }
+
+  /// Removes certain classes and leak types from skip lists in [settings].
+  ///
+  /// Items in [classes] will be removed from all skip lists.
+  static void track({
```
kenzieschmoll, line 120: r/track/`trackLeaks` to be consistent with opposite `allowLeaks`

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+      notGCed: notGCed,
+      classes: classes,
+    );
+  }
+}
+
+/// Leak tracking settings for tests.
+///
+/// Should be instantiated using static methods of [LeakTrackingForTests].
+class LeakTrackingForTestsSettings {
```
kenzieschmoll, line 136: should this be private?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  /// If true, leak tracking is paused.
+  final bool paused;
+
+  /// If true, tests will fail on leaks.
+  ///
+  /// Set to true to test that tests collect expected leaks.
+  final bool failOnLeaks;
+
+  /// Callback to invoke if test collected leaks, before failing if [failOnLeaks] is true.
```
kenzieschmoll, line 172: wording nit: Callback to invoke before the test fails when [failOnLeaks] is true and leaks were found.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+  final LeaksCallback onLeaks;
```
kenzieschmoll, line 173: nit: onLeaksFound? onLeaksCollected?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  /// Leaks to skip.
+  final LeakSkipLists leakSkipLists;
```
kenzieschmoll, line 176: allowedLeaks

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  /// Defines which disgnostics information to collect.
```
kenzieschmoll, line 178: sp: diagnostic

kenzieschmoll (commented):
Some comments around naming that might make these classes / method names more intuitive.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
```diff
@@ -38,6 +40,104 @@ class Switches {
   bool get isObjectTrackingDisabled => disableNotDisposed && disableNotGCed;
 }
 
+/// List of classes to skip during leak tracking.
+class LeakSkipList {
```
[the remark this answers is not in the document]
polina-c: In general I agree allowed is a standard.

In this concrete case when I worked with leaks, I noticed the word 'allow', that is used now to exempt leaks in FF, is confusing: do we want to allow leaks to happen or allow exception about the leak to happen.

The word 'skip' is preciser in this situation, as it means we do not want to notice the leak.

How does it sound?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: It cannot be instantiated by user, but user can see the instance and create other instances out of it.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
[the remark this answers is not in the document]
polina-c: The same relates to `remove` and `contains`. 
Remove from leaks we want to track or remove from allow/skip list?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+
+    return settings.copyWith(
+      leakSkipLists: LeakSkipLists(
+        notGCed: settings.leakSkipLists.notGCed.merge(
+          LeakSkipList(
+            byClass: addClassesToMap(notGCed, classes),
+            skipAll: allNotGCed,
+          ),
+        ),
+        notDisposed: settings.leakSkipLists.notGCed.merge(
```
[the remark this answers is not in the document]
polina-c: yes, thanks, added test coverage

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,183 @@
...
+      allNotGCed: allNotGCed,
+      classes: classes,
+    );
+  }
+
+  /// Returns copy of [settings] with extended skip lists.
+  ///
+  /// In the result the skip limit for a class is maximum of two original skip limits.
+  /// Items in [classes] will be added to all skip lists.
+  static LeakTrackingForTestsSettings withSkipped({
```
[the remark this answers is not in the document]
polina-c: added `copy` and `Leaks`

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: added `Leaks`

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,198 @@
...
+
+  /// Updates [settings] to pause leak tracking.
+  static void pause() => settings = paused();
+
+  /// Updates [settings] to start leak tracking.
+  static void start() => settings = started();
+
+  /// Creates a copy of current settings with [paused] set to true.
+  static LeakTrackingForTestsSettings paused() =>
+      settings.copyWith(paused: true);
```
polina-c, line 24: I did not use name `withPaused` and `withStarted` because this method will be used massively, to exempt individual tests from leak tracking.

After merging this PR, I will update flutter/test, to have 8% of tests that are not leak free yet looking like this: 

```
    testWidgets('initialTimerDuration falls within limit',
    experimentalLeakTracking: LeakTrackingForTests.paused(),
    (WidgetTester tester) async {
        ...
```

withPaused looks less readable.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
```diff
@@ -109,17 +109,17 @@ class LeakSkipList {
 }
 
 /// Skip lists for each type of leak.
```
kenzieschmoll, line 111: nit: update dart-doc. "The total set of skipped leaks for both [nonGCed] and [notDisposed] leaks."

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
```diff
@@ -40,13 +40,13 @@ class Switches {
   bool get isObjectTrackingDisabled => disableNotDisposed && disableNotGCed;
 }
 
-/// List of classes to skip during leak tracking.
-class LeakSkipList {
-  const LeakSkipList({this.byClass = const {}, this.skipAll = false});
+/// Set of classes to skip during leak tracking.
+class LeakSkipSet {
```
kenzieschmoll, line 44: nit: SkippedLeakSet to be more consistent with `SkippedLeaks`?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
kenzieschmoll: wdyt about replacing "paused" with "ignore" or "skip" and replacing "started" with "track"? Pause and start are confusing to me because this implies some sort of live and modifiable state, whereas it is actually a static setting defined by code. This would also be consistent with the "skip" and "track" terminology used elsewhere which is nice.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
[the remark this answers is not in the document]
kenzieschmoll: I agree this isn't a simple feature set to name... reviewed your changes and what you have looks good. Just a few more nits.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
goderbauer, line 1: Should this file be exported by `lib/leak_tracker_flutter_testing.dart` or is it just for package-internal use?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
goderbauer: +1 to this

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,198 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:leak_tracker/leak_tracker.dart';
+import 'package:meta/meta.dart';
+
+void _emptyLeakHandler(Leaks leaks) {}
+
+/// Leak tracking settings and helper methods for tests.
+abstract class LeakTrackingForTests {
```
goderbauer, line 11: It is not entirely clear to me how this class is supposed to be used (some docs (incl. usage examples)) would help).

in the example given in the comment it appears to be used to configure a `testWidgets` test case - I am confused that it contains mutable state, though. Can you change the test configuration while a test is running? Why is that necessary?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,198 @@
...
+  /// Current configuration for leak tracking.
+  static LeakTrackingForTestsSettings settings =
```
goderbauer, line 13: Why is this a global state singleton? Shouldn't I be able to configure this for each test separately? And should this field be user-writable instead of requireing users to go through the defined methods below?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,198 @@
...
+      const LeakTrackingForTestsSettings();
+
+  /// Updates [settings] to pause leak tracking.
+  static void pause() => settings = paused();
+
+  /// Updates [settings] to start leak tracking.
+  static void start() => settings = started();
+
+  /// Creates a copy of current settings with [paused] set to true.
+  static LeakTrackingForTestsSettings paused() =>
```
goderbauer, line 23: This API is a bit confusing. It is not clear to me as a user what the difference between calling `pause` and `paused` is supposed to be.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
goderbauer, line 11: Why does this class with only static members exist in the first place? https://dart.dev/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: good idea

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
```diff
@@ -38,6 +40,104 @@ class Switches {
   bool get isObjectTrackingDisabled => disableNotDisposed && disableNotGCed;
 }
 
+/// Set of classes to skip during leak tracking.
+class LeakSkipSet {
+  const LeakSkipSet({this.byClass = const {}, this.skipAll = false});
```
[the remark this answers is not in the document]
polina-c: Added comment.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: Added comment for the class.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: Comment for the class may help.
I do not want users to create the settings from scratch, because I want defaults for package to affect all the test. So that if you set leak tracking to paused for the entire package, it would stop all the tests, except those that explicitly set paused=false, in spite of default.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: Comment for the class should help.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: The rule states 'AVOID', not 'DON'T'.

This rule is violated across Dart and Flutter many times (https://api.dart.dev/stable/3.0.6/dart-developer/Service/Service.html).

I think we need to convince language team to provide a way to group static methods, so that we do not use abstract classes for these purposes.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: And `paused` is renamed to `ignore` now.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -7,24 +7,48 @@ import 'package:meta/meta.dart';
...
   static LeakTrackingForTestsSettings settings =
       const LeakTrackingForTestsSettings();
 
   /// Updates [settings] to pause leak tracking.
-  static void pause() => settings = paused();
+  static void pause() => settings = ignore();
 
   /// Updates [settings] to start leak tracking.
-  static void start() => settings = started();
+  static void start() => settings = track();
```
kenzieschmoll, line 44: these need to be renamed to ignore and track

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -7,24 +7,48 @@ import 'package:meta/meta.dart';
...
 
-  /// Creates a copy of current settings with [paused] set to true.
-  static LeakTrackingForTestsSettings paused() =>
+  /// Creates a copy of current settings with [LeakTrackingForTestsSettings.paused] set to true.
+  static LeakTrackingForTestsSettings ignore() =>
       settings.copyWith(paused: true);
```
kenzieschmoll, line 48: should these usages of paused be changed to ignore?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -7,24 +7,48 @@ import 'package:meta/meta.dart';
...
 
-  /// Creates a copy of current settings with [paused] set to false.
-  static LeakTrackingForTestsSettings started() =>
+  /// Creates a copy of current settings with [LeakTrackingForTestsSettings.paused] set to false.
+  static LeakTrackingForTestsSettings track() =>
       settings.copyWith(paused: false);
```
kenzieschmoll, line 52: same comment here

kenzieschmoll (approved):
lgtm from me after cleaning up lingering "pause" names.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: There are two static methods: 
ignore() - returns instance of settings with paused set to true
pause - updates current settings with paused set to true

It looks meaningful for me, because first one is intended to skip just one test, while second one stops leak tracking for any test, that uses default.

We can think about other names, but we need two different names to avoid collision.

kenzieschmoll: Based on the comment thread here, https://github.com/dart-lang/leak_tracker/pull/152#discussion_r1361003657, I thought we agreed to replace `pause` and `start` with `ignore` and `track`? With all four terminologies in the mix this may be difficult for a user to understand the API. It is confusing to have 4 terms instead of just 2 that are consistent throughout.

polina-c: Okay, I see, you suggest to replace it everywhere:

SkippedLeaks (and all its members) -> IgnoredLeaks 
LeakTrackingForTestsSettings.pause -> LeakTrackingForTestsSettings.ignore
Static methods `ignore` (update settings) and `ignored` (return copy of settings with ignore = true)

kenzieschmoll: > SkippedLeaks (and all its members) -> IgnoredLeaks

This wasn't part of my original suggestion, but it could make sense to keep terminology consistent throughout this API. @goderbauer do you have an opinion on naming here?

My comment was regarding current uses of `pause` and `start`. So
`LeakTrackingForTestsSettings.pause -> LeakTrackingForTestsSettings.ignore` and 
`LeakTrackingForTestsSettings.start -> LeakTrackingForTestsSettings.track`.

Also changing the `copyWith` parameter name `paused` to `ignored` or maybe `ignoreAll` would be better. Same for the class member `paused`.

polina-c: Updated everything to 'ignore'. @goderbauer , I am still open for other suggestions. Naming is important and it is better to make it right now.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,222 @@
...
+/// Use methods that return adjusted [LeakTrackingForTests.settings]
+/// to customize default for an individual test:
+///
+/// ```dart
+/// testWidgets(
+///     'initialTimerDuration falls within limit',
+///     leakTracking: LeakTrackingForTests.withIgnoredLeaks(allNotGCed: true),
+///     (WidgetTester tester) async {
+///       ...
+/// ```
```
goderbauer, line 29: The API and the way this class can be used feel super complex. Are all the different ways in which this class can be used necessary? What are the use cases that this enables?

Naively, I would have expected that we need to support two use cases:
1) Setting a default setting for all `testWidgets` that don't configure a local setting.
2) On a per test basis, `testsWidgets` should be able to override the defaults.

For that I would have expected something simple like: I can instantiate an immutable `LeakTrackingSettings` object. It's either configured by setting arguments in the `LeakTrackingSettings` constructor or I can invoke named constructors for some well known use cases (e.g. to track all leaks). I can then either pass that constructed object to `testWidgets` to configure something on a test-by-test basis, or I assign that object to a static (e.g. `LeakTrackingSettings.defaults`) to configure the defaults for all tests (e.g. from `flutter_test_config.dart`). 

The immutable `LeakTrackingSettings` object could have a copyWith method to create a setting derived from another. For example, if I just want to change one thing for a given test from the default settings I would do something like:

```dart
testWidgets('foo', leakTracking: LeakTrackingSettings.defaults.copyWith(barSetting: false), ...);
```

There could even be convenience copy methods for well-known use cases.

This would give `LeakTrackingSettings` a single, easy-to-understand API and it would make it explicit from what setting something is derived from.

Note: I am using `LeakTrackingSettings` as a placeholder name in the above. It could be called something else.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,222 @@
...
+///
+/// Update for [LeakTrackingForTests.settings] when a test is already running
+/// will throw exception.
```
goderbauer, line 32: This doesn't seem to match the implementation since settings is just a field?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
```diff
@@ -0,0 +1,222 @@
...
+      notDisposed: notDisposed,
+      notGCed: notGCed,
+      classes: classes,
+    );
+  }
+}
+
+/// Leak tracking settings for tests.
+///
+/// Should be instantiated using static methods of [LeakTrackingForTests].
```
goderbauer, line 173: Why is that a separate class then instead of just defining these on LeakTrackingForTestsSettings directly?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
goderbauer: It feels like LeakTrackingForTestsSettings and LeakTrackingForTests should just be one class then? Why do the statics need to be separated out?

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: For now nothing matches implementation, because leak tracker does not use the settings yet.
It is work in progress.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
```diff
@@ -38,6 +40,110 @@ class Switches {
...
+  /// Removes the classes from ignore lists.
+  IgnoredLeaksSet track(List<String> list) {
+    if (list.isEmpty) return this;
+    final map = {...byClass};
+    list.forEach(map.remove);
+    return copyWith(byClass: map);
+  }
+
+  /// Returns true if the class should be ignoreped.
+  bool isIgnored(String className) {
```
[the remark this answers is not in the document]
polina-c: Added comment.

pkgs/leak_tracker_flutter_testing/lib/src/leak_tracking_for_tests.dart
polina-c: Updated comment.

pkgs/leak_tracker/lib/src/leak_tracking/_primitives/model.dart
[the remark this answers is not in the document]
goderbauer: (Not sure what use this API has if it doesn't consider the limit 🤷)

pkgs/leak_tracker_flutter_testing/lib/src/leak_testing.dart
```diff
@@ -0,0 +1,187 @@
...
+/// If you update the settings for a group, remember the original value to a local variable
+/// and restore it in `tearDownAll` for the group.
+///
+/// Use methods that return adjusted [LeakTesting.settings]
+/// to customize default for an individual test:
+///
+/// ```dart
+/// testWidgets(
+///     'initialTimerDuration falls within limit',
+///     leakTracking: LeakTesting.withIgnoredAll(),
```
goderbauer, line 26:
```suggestion
///     leakTracking: LeakTesting.settings.withIgnoredAll(),
```

pkgs/leak_tracker_flutter_testing/test/tests/leak_tracking_for_tests_test.dart
```diff
@@ -0,0 +1,215 @@
...
+            theSettings.ignoredLeaks.isIgnored(theClass, leakType: leakType),
+      )
+      .any((skipped) => !skipped);
+  return classesSkipped && othersTracked;
+}
+
+void main() {
+  setUp(() {
+    LeakTesting.settings = const LeakTesting();
+  });
```
goderbauer, line 63: Could you do the following and then remove `@visibleForTesting` from the LeakTesting constructor and make it actually private to underline that nobody should be instantiating it directly?

```suggestion
  late LeakTesting defaults;

  setUp(() {
    defaults = LeakTesting.settings;
  });
  tearDown(() {
    LeakTesting.settings = defaults;
  });
```

goderbauer (approved):
LGTM

PR #168, opened 2023-10-31
Reorganize code to simplify dependencies.

polina-c:
Prerequisite: https://github.com/flutter/flutter/pull/137633

1. Remove dependency on leak_tracker_testing in leak_tracker.
2. Move LeakTesting from leak_tracker_flutter_testing to leak_tracker, so that flutter_test that needs it, could avoid circular dependency.
3. Rename autosnapshotting to autosnapshotting_example, to make error messages for version resoultion clearer.


pkgs/leak_tracker/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: leak_tracker
-version: 9.0.12
+version: 9.0.13
```
bkonyi, line 2: Did you already release 9.0.12?

pkgs/leak_tracker_flutter_testing/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.0.11
+
+* Move LeakTesting out of this package to leak_tracker.
```
bkonyi, line 3: Is this a breaking change?

bkonyi (approved):

PR #169, opened 2023-11-01
Separate memory_usage.

polina-c:
Contributes to https://github.com/flutter/flutter/issues/135856


README.md
```diff
@@ -9,9 +9,23 @@ This is a framework for detecting memory issues in Dart and Flutter applications
...
+| [leak_tracker](pkgs/leak_tracker/) | (work in progress, used by flutter_test) A framework for detecting memory issues for Dart and Flutter applications. | [![pub package](https://img.shields.io/pub/v/leak_tracker.svg)](https://pub.dev/packages/leak_tracker) |
+| [leak_tracker_testing](pkgs/leak_tracker_testing/) | (work in progress, used by flutter_test) Leak tracking helpers intended for usage in Dart and Flutter tests. | [![pub package](https://img.shields.io/pub/v/leak_tracker_testing.svg)](https://pub.dev/packages/leak_tracker_testing) |
+| [leak_tracker_flutter_testing](pkgs/leak_tracker_flutter_testing/) | Internal package to test leak tracker with Flutter. | [![pub package](https://img.shields.io/pub/v/leak_tracker_flutter_testing.svg)](https://pub.dev/packages/leak_tracker_flutter_testing) |
+| [memory_usage](pkgs/memory_usage/) | (experimental, used by Dart SDK) Functions to listen memory usage events and to auto-snapshot memory in case of overuse. | [![pub package](https://img.shields.io/pub/v/memory_usage.svg)](https://pub.dev/packages/memory_usage) |
```
devoncarew, line 15: Perhaps:

```suggestion
| [memory_usage](pkgs/memory_usage/) | (experimental) Functionality to listen to memory usage events and to auto-snapshot memory in case of overuse. | [![pub package](https://img.shields.io/pub/v/memory_usage.svg)](https://pub.dev/packages/memory_usage) |
```

examples/autosnapshotting/pubspec.yaml
```diff
@@ -11,9 +9,8 @@ dependencies:
     sdk: flutter
   cupertino_icons: ^1.0.2
   intl: ^0.18.1
-  # The version is `any` because it is defined by integration_test, that depends on flutter_test,
-  # whish depends on leak_tracker.
-  leak_tracker: any
+  # The version is `any` because it is defined Dart SDK.
```
devoncarew, line 12: I'd remove this comment (it's `any` because it's consumed from the same repo).

pkgs/memory_usage/CHANGELOG.md
```diff
@@ -0,0 +1,167 @@
+# 9.0.13
```
devoncarew, line 1: I would remove the contents here with one changelog entry for `0.1.0`. The items in the changelog entry can be just a list of user facing features.

pkgs/memory_usage/README.md
```diff
@@ -0,0 +1,23 @@
+[![pub package](https://img.shields.io/pub/v/leak_tracker.svg)](https://pub.dev/packages/leak_tracker)
+[![package publisher](https://img.shields.io/pub/publisher/leak_tracker.svg)](https://pub.dev/packages/leak_tracker/publisher)
+
+Coming soon! See https://github.com/flutter/devtools/issues/3951 and https://github.com/flutter/devtools/issues/5606.
```
devoncarew, line 4: If these links are more related to package:leak_tracker than package:memory_usage, I'd remove them.

pkgs/memory_usage/README.md
```diff
@@ -0,0 +1,23 @@
+[![pub package](https://img.shields.io/pub/v/leak_tracker.svg)](https://pub.dev/packages/leak_tracker)
```
devoncarew, line 1: The things in this file which refer to package:leak_tracker (not the repo) should be updated to instead preference package:memory_usage (i.e., the pub.dev references).

pkgs/memory_usage/README.md
```diff
@@ -0,0 +1,23 @@
...
+For packages in the labs.dart.dev publisher we generally plan to either graduate
+the package into a supported publisher (dart.dev, tools.dart.dev) after a period
+of feedback and iteration, or discontinue the package. These packages have a
+much higher expected rate of API and breaking changes.
+
+Your feedback is valuable and will help us evolve this package. For general
+feedback, suggestions, and comments, please file an issue in the
+[bug tracker](https://github.com/dart-lang/leak_tracker/issues).
+
+## What is this?
```
devoncarew, line 21: Keep in mind that visitors to the package on pub.dev will mainly see this readme. It's worth expanding the documentation here even if its duplicative of other docs (those more complete docs can then be referenced here).

Minimally, I would retain the 'memory usage tracking' section from https://pub.dev/packages/leak_tracker.

pkgs/memory_usage/example/main.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:memory_usage/memory_usage.dart';
```
devoncarew, line 1: This file needs our std copyright header.

pkgs/memory_usage/example/pubspec.yaml
```diff
@@ -0,0 +1,13 @@
...
+publish_to: none
+
+environment:
+  sdk: ^3.0.0
+
+dependencies:
+  memory_usage: ^0.1.0
+
+dependency_overrides:
+  flutter_lints: ^3.0.0
```
devoncarew, line 11: You'll want to move the linting dependency into a `dev_dependency` section.

pkgs/memory_usage/lib/src/_util.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2023 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+extension SizeConversion on int {
+  int mbToBytes() => this * 1024 * 1024;
```
devoncarew, line 6: a nit, since this is existing code, but this should likely be a getter.

pkgs/memory_usage/analysis_options.yaml
```diff
@@ -0,0 +1,143 @@
+include: package:flutter_lints/flutter.yaml
```
devoncarew, line 1: Since this is not a flutter package, I would highly recommend moving to either package:lints/recommended.yaml (what we recommend to dart users) or package:dart_flutter_team_lints/analysis_options.yaml (what we use for many package in dart_lang). The 2nd will be more restrictive, so the 1st may be easier to move to. Either one will replace the contents of this file w/ a simple include which will be a nice benefit.

devoncarew (commented):
Looks comprehensive! I think most of my comments on the PR are housekeeping in nature (due to the code moves).

I see the new code in pkgs/memory_usage but not the removal from pkgs/leak_tracker. I don't know that its critical to do in this PR as long as we circle back to it (it would be a major version change for that package).

pkgs/memory_usage/analysis_options.yaml
polina-c: For now I am copy-pasting this across packages. Created issue to fix it: https://github.com/dart-lang/leak_tracker/issues/175

Thanks.

pkgs/memory_usage/lib/src/_util.dart
polina-c: I prefer not to change copied code as it is proven to work and it is not test covered in leak_tracker that much.

polina-c:
> Looks comprehensive! I think most of my comments on the PR are housekeeping in nature (due to the code moves).
> 
> I see the new code in pkgs/memory_usage but not the removal from pkgs/leak_tracker. I don't know that its critical to do in this PR as long as we circle back to it (it would be a major version change for that package).

Yes, after updating analyzer.

pkgs/leak_tracker/pubspec.yaml
```diff
@@ -6,6 +6,9 @@ repository: https://github.com/dart-lang/leak_tracker/tree/main/pkgs/leak_tracke
 environment:
   sdk: '>=3.1.2 <4.0.0'
 
+# All dependencies here will become Flutter transitive dependencies.
```
devoncarew, line 9: 👍

pkgs/memory_usage/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+This package enables [memory usage tracking and auto-snapshotting](https://github.com/dart-lang/leak_tracker/blob/main/doc/USAGE.md).
+
+## Usage
+
+Use the function trackMemoryUsage to configure usage events and auto-snapshotting in your Dart or Flutter application.
```
devoncarew, line 27:
```suggestion
Use the function `trackMemoryUsage` to configure usage events and auto-snapshotting in your Dart or Flutter application.
```

pkgs/memory_usage/example/pubspec.yaml
devoncarew: You'll need to create a new section here - `dev_dependency`, have flutter_lints specified there, and retain the `dependency_overrides` section for use by the `memory_usage` dep.

pkgs/memory_usage/lib/src/model.dart
```diff
@@ -0,0 +1,188 @@
...
+
+  /// A callback that is called when a snapshot is taken.
+  final SnapshotCallback? onSnapshot;
+
+  /// The absolute path to the [directory].
+  String get directoryAbsolute => path.absolute(directory);
+
+  @override
+  String toString() {
+    final formatter = NumberFormat('#,###,000');
```
devoncarew, line 180: As a note to self, if the only usage of package:intl from this package is for number formatting to toString methods (I assume just used when debugging?), we should be able to prune the dep on `intl` w/o much difficulty.

devoncarew: (No real feedback for this PR however as this is existing code)

pkgs/memory_usage/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: memory_usage
+version: 0.1.0
```
devoncarew, line 2: I'd like to follow up this PR w/ one or two changes (if I don't end up doing that in the next day or two we can remove this wip and publish).

```suggestion
version: 0.1.0-wip
```

devoncarew:
> > I see the new code in pkgs/memory_usage but not the removal from pkgs/leak_tracker. I don't know that its critical to do in this PR as long as we circle back to it (it would be a major version change for that package).
> 
> Yes, after updating analyzer.

It's not critical for this PR, but updates like that will happen atomically in the SDK. So, it'll see the old dart-lang/leak_tracker repo dep, and the code in package:leak_tracker, or the new dart-lang/leak_tracker dep, and the code in package:memory_usage. The CL that updates the repo dep to the latest will also update the imports using the memory tracking code in the analysis server.

pkgs/memory_usage/example/pubspec.yaml
polina-c: I do not know how it continue squeezing to the PR. Fixed.

pkgs/memory_usage/lib/src/model.dart
polina-c: removed. thanks

polina-c:
> > > I see the new code in pkgs/memory_usage but not the removal from pkgs/leak_tracker. I don't know that its critical to do in this PR as long as we circle back to it (it would be a major version change for that package).
> > 
> > 
> > Yes, after updating analyzer.
> 
> It's not critical for this PR, but updates like that will happen atomically in the SDK. So, it'll see the old dart-lang/leak_tracker repo dep, and the code in package:leak_tracker, or the new dart-lang/leak_tracker dep, and the code in package:memory_usage. The CL that updates the repo dep to the latest will also update the imports using the memory tracking code in the analysis server.

Good to know. I prefer to be extra careful with such migrations :)

pkgs/leak_tracker/lib/src/usage_tracking/model.dart
```diff
@@ -174,15 +167,4 @@ class AutoSnapshottingConfig {
 
   /// The absolute path to the [directory].
   String get directoryAbsolute => path.absolute(directory);
-
-  @override
-  String toString() {
-    final formatter = NumberFormat('#,###,000');
```
devoncarew, line 180: As a final comment, I have no concerns whatsoever if we keep these toString() methods, we'd just likely want a slightly less sophisticated way of formatting numbers.

devoncarew (approved):
lgtm!

PR #176, opened 2023-11-03
Add arguments to withTracked.

polina-c:


pkgs/leak_tracker/lib/src/leak_tracking/leak_testing.dart
```diff
@@ -138,12 +138,24 @@ class LeakTesting {
     List<String> notGCed = const [],
     List<String> notDisposed = const [],
     List<String> classes = const [],
+    bool allNotGCed = false,
+    bool allNotDisposed = false,
```
kenzieschmoll, line 142: I don't see anywhere in this PR where these parameters are actually used? We should add a tests that exercises the `withTracked` method with the new parameters.

kenzieschmoll (approved):

PR #180, opened 2023-11-03
use package:lints/recommended.yaml generally

devoncarew:
- use package:lints/recommended.yaml generally
- borrow a few lints from package:dart_flutter_team_lints to catch additional issues (moving to the whole set would let us have smaller analysis options configs here; we'd need to address a few more findings)

Note there there is one existing error - a doc reference - which I wasn't sure what to update it to.

---

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>


devoncarew:
```
Analyzing leak_tracker_flutter_testing...                       

   info • The referenced name isn't visible in scope • lib/src/test_widgets.dart:78:6 • comment_references

```

Screenshot 2023-11-03 at 2 09 33 PM

I'm not sure what that should be updated to -

polina-c:
> ```
> Analyzing leak_tracker_flutter_testing...                       
> 
>    info • The referenced name isn't visible in scope • lib/src/test_widgets.dart:78:6 • comment_references
> ```
> 
> Screenshot 2023-11-03 at 2 09 33 PM
> I'm not sure what that should be updated to -

I will clean up this code very soon. This can be just deleted.

pkgs/leak_tracker/lib/src/leak_tracking/_baseliner.dart
```diff
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// ignore_for_file: avoid_print
```
devoncarew, line 5: @polina-c - there are two `print` statements in this file; I suspect they're bugs (should be removed? the output routed to a file / to devtools?).

polina-c: They are not bugs. It is functionality, that will be used for tests. Tests are expected to  output the information.

devoncarew: To stdout? What if users call that code? We'll produce random lines to their console.

polina-c: This will output will happen if only user asked this to be outputed. Intentionally. The tests will output performance metrics.

devoncarew: OK, understood.

FYI that we are no longer reporting `avoid_print` (print uses) in this file in any case, as the linting is not based on flutter/lints.

polina-c (approved):

PR #186, opened 2023-11-23
chore: refactor code style

LucasXu0:
improve the code style and make the code more clear

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

### DIFF
--- a/pkgs/leak_tracker/lib/src/shared/shared_model.dart
+++ b/pkgs/leak_tracker/lib/src/shared/shared_model.dart
@@ -23,8 +23,7 @@ enum LeakType {
   /// Disposed and garbage collected later than expected.
   gcedLate;
 
-  static LeakType byName(String name) =>
-      LeakType.values.firstWhere((e) => e.name == name);
+  static LeakType byName(String name) => LeakType.values.byName(name);
 }
 
 /// Names for json fields.


pkgs/leak_tracker/lib/src/shared/shared_model.dart
```diff
@@ -23,8 +23,7 @@ enum LeakType {
   /// Disposed and garbage collected later than expected.
   gcedLate;
 
-  static LeakType byName(String name) =>
-      LeakType.values.firstWhere((e) => e.name == name);
+  static LeakType byName(String name) => LeakType.values.byName(name);
 }
```
polina-c, line 27: I like this.

pkgs/leak_tracker/lib/src/leak_tracking/_object_record.dart
```diff
@@ -97,19 +97,14 @@ class ObjectRecord {
...
-    if (theContext == null) {
-      context = addedContext;
-      return;
-    }
-    theContext.addAll(addedContext);
+    context ??= {};
+    context!.addAll(addedContext);
   }
 
   LeakReport toLeakReport() => LeakReport(
```
polina-c, line 110: The intent of the original code to avoid `!` and to use null-safety.
The new code has '!' and thus moves null-safety validation from compiler to developer, making it error prone.
Can you revert changes in this files?

LucasXu0: sure.

LucasXu0: I used the ! syntax because the context must not be null in this case.

How about this change? I will submit it if it's acceptable.

```dart
  void setContext(String key, Object value) {
    context ??= {};
    context?[key] = value;
  }

  void mergeContext(Map<String, dynamic>? addedContext) {
    if (addedContext == null) return;
    context ??= {};
    context?.addAll(addedContext);
  }
```

polina-c: Looks good. Thank you.

polina-c (approved):

polina-c:
Merged the change. Feel free to create new PR with others.

PR #188, opened 2023-12-01
Add API to integrate with testWidgets.

polina-c:


pkgs/leak_tracker_flutter_testing/lib/src/testing.dart
```diff
@@ -0,0 +1,99 @@
...
+///
+/// Used to test leak tracking functionality.
+LeaksCallback collectedLeaksReporter =
+    (Leaks leaks) => expect(leaks, isLeakFree);
+
+void _dispatchFlutterEventToLeakTracker(ObjectEvent event) {
+  return LeakTracking.dispatchObjectEvent(event.toMap());
+}
+
+bool _notSupportedWarningPrinted = false;
```
CoderDake, line 67: variables should probably be together at the top or bottom of the file.

CoderDake (approved):

pkgs/leak_tracker_flutter_testing/lib/src/testing.dart
polina-c: This exact variable is used only in _checkPlatformAndMayBePrintWarning, and is private. It is easier to read the method if the variable is closeby.

PR #190, opened 2023-12-08
fix: add missing on leak callback

LucasXu0:

The onLeaks callback did not get invoked.


polina-c:
This code is under refactoring and will be removed.
I will merge it in case you need it to track leaks now.

See the plan here: https://github.com/flutter/flutter/issues/135856

polina-c (approved):

polina-c:
Feel free to add PR to increase version so that you can take dependency on the fixed leak_tracker.

LucasXu0:
> Feel free to add PR to increase version so that you can take dependency on the fixed leak_tracker.

Sure!

> This code is under refactoring and will be removed. I will merge it in case you need it to track leaks now.
> 
> See the plan here: [flutter/flutter#135856](https://github.com/flutter/flutter/issues/135856)

Thanks for the information.

PR #194, opened 2023-12-15
Remove declaration of testWidgetsWithLeakTracking.

polina-c:
We need it to move dependency on flutter_test from dependencies to dev_dependencies, to avoid circular reference after integration with testWidgets.


kenzieschmoll:
Are all the test removed in this PR going to be added back somewhere else?

polina-c:
> Are all the test removed in this PR going to be added back somewhere else?

The tests are recreated in this PR: https://github.com/flutter/flutter/pull/138057

kenzieschmoll (approved):

PR #199, opened 2023-12-30
Enhance scripting.

polina-c:


tool/analyze.sh
```diff
@@ -9,28 +9,31 @@
 # Fast fail the script on failures.
 set -ex
 
-sh ./tool/pub_get.sh
+# The directory that this script is located in.
+TOOL_DIR=`dirname "$0"`
```
CoderDake, line 13:
```suggestion
TOOL_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
If you get the full path of the script then you  can avoid using `cd -` below.

The same applies to the other files in this change.

tool/analyze.sh
```diff
@@ -9,28 +9,31 @@
...
+cd $TOOL_DIR/../pkgs/leak_tracker
 dart analyze --fatal-infos
 cd -
 
-cd pkgs/leak_tracker_flutter_testing
+cd $TOOL_DIR/../pkgs/leak_tracker_flutter_testing
 flutter analyze --fatal-infos
 cd -
 
 cd pkgs/leak_tracker_testing
```
CoderDake, line 33: Looks like some of the pkg cds are missing $TOOL_DIR?

tool/analyze.sh
CoderDake: I got that one liner from: https://stackoverflow.com/a/246128

polina-c: I use tool dir to navigate to other directories in order to run flutter commands from project root. How path to the script will help me?

CoderDake: it will just make it so you don't have to  cd - each time. you can just jump directly to the next directory you are working in, i.e.:
```
cd $TOOL_DIR/../examples/autosnapshotting
flutter analyze  --fatal-infos

cd $TOOL_DIR/../examples/leak_tracking
flutter analyze --fatal-infos

cd $TOOL_DIR/../pkgs/leak_tracker
dart analyze --fatal-infos
```

CoderDake: since it is the full path, there is no need to use the `cd -` :)

polina-c:
Closed in favor of https://github.com/dart-lang/leak_tracker/pull/204

PR #203, opened 2024-01-02
Handle deprecation in Flutter.

polina-c:


polina-c:
@devoncarew , can you review?

examples/leak_tracking/lib/main.dart
```diff
@@ -4,7 +4,7 @@ import 'package:leak_tracker/leak_tracker.dart';
 
 void main() {
   LeakTracking.start();
-  MemoryAllocations.instance.addListener(
+  FlutterMemoryAllocations.instance.addListener(
```
devoncarew, line 7: It looks like this typedef comes from the package:flutter import; when first reading this I had assumed this was from package:leak_tracker. Perhaps clarify this (via comments) in this example?

pkgs/leak_tracker_flutter_testing/lib/src/matchers.dart
```diff
@@ -21,9 +21,9 @@ Future<List<ObjectEvent>> memoryEvents(
     }
   }
 
-  MemoryAllocations.instance.addListener(listener);
+  FlutterMemoryAllocations.instance.addListener(listener);
```
devoncarew, line 24: When was `FlutterMemoryAllocations` added to package:flutter? Do we need to update the min. sdk constraint for this package (in order to specify a min. sdk that contains this typedef)?

devoncarew (approved):

PR #208, opened 2024-01-18
Document perf issues to ignore test helpers.

polina-c:


pkgs/leak_tracker_testing/lib/src/leak_testing.dart
```diff
@@ -121,6 +121,10 @@ class LeakTesting {
   /// In the result the ignored limit for a class is the
   /// maximum of two original ignored limits.
   /// Items in [classes] will be added to all ignore lists.
+  ///
+  /// Setting [createdByTestHelpers] to true may cause significant
+  /// performance impact on the test run, caused by conversion of
+  /// creation call stack to String.
```
kenzieschmoll, line 127: is there any way to improve the performance here instead of just documenting that it is poor?

kenzieschmoll (approved):
dart doc lgtm. One general question though.

pkgs/leak_tracker_testing/lib/src/leak_testing.dart
polina-c: I did not find. We can brainstorm it together :)

Issue #118, opened 2023-08-07
How leak_tracker should handle double disposal and double creation?


polina-c:
Right now leak_tracker ignores cases of double disposal or double start of tracking.
It is possible for leak tracker to fail or give warning or include into leak report.

This issue is created to discuss options to implement it.

polina-c:
Not all users of leak_tracker want leak_tracker to fail on double dispatch of creation or disposal.
So, there should be option to configure leak_tracker to fail or not to fail in such cases.

PR #226, opened 2024-03-31
Fix handling of double tracking for an object.

polina-c:
Dispatch of creation and disposal should be idempotent. Leak tracker should not fail or track the same object twice.

Issue https://github.com/dart-lang/leak_tracker/issues/118 suggests more advanced handling.


pkgs/leak_tracker/lib/src/leak_tracking/_object_record.dart
```diff
@@ -48,10 +48,10 @@ class ObjectRecord {
   int? _disposalGcCount;
 
   void setDisposed(int gcTime, DateTime time) {
-    // TODO(polina-c): handle double disposal in a better way
-    // https://github.com/dart-lang/leak_tracker/issues/118
-    // Noop if object is already disposed.
-    if (_disposalGcCount != null) return;
+    if (_disposalGcCount != null) {
+      // It is not responsibility of leak tracker to check for double disposal.
```
jacob314, line 52: More of a CL description would help clarify what this CL is doing. Is it handling tracking or just ignoring the double dispose problem? Would expect issue/118 would be closed if the design decision is to not handle the case.

jacob314 (approved):
lgtm

pkgs/leak_tracker/lib/src/leak_tracking/_object_record.dart
polina-c: And added description.

PR #246, opened 2024-09-10
Recommend to opt out hacky tests on the page that is linked in leak tracking error.

polina-c:


doc/leak_tracking/TROUBLESHOOT.md
```diff
@@ -69,7 +69,16 @@ to happen after the rendering cycle completes.
 If this is a case, `imageCache.clear()` needs to happen as last statement of the test,
 instead of in tear down, to allow the cycles to happen.
 
-### 4. The test throws flutter exception
+### 4. The test is hacky
```
Piinks, line 72: Following up on request from https://github.com/flutter/flutter/pull/154481
IMO, hacky tests should generally be avoided. :)
Does this scenario need to be documented then? It feels like a blank check that anyone could say, well let's just skip testing for leaks since this test is a hack. That feels like a test that is not very valuable.

polina-c: There can be valuable tests that perform important verifications, using "wrong" mechanisms that are restricted in production. Depending on situation and cost we use mocks, stubs, or "hacks".
In such tests we can say that objects are leaking by design.

